PR #38, opened 2015-01-21
TraitsUI Undo integration

corranwebster:
This PR provides a simple `Command` that resets/sets a single trait value on undo/redo, plus a Traits UI `Handler` subclass that overrides the `setattr` method to instead push one of these commands onto a command stack (which will, when the `do` method is called, actually set the trait).

This means that a traits view-based UI can easily be adapted to use an application's undo/redo stack by simply using the new handler.

### DIFF
--- a/apptools/undo/traitsui/trait_set_command.py
+++ b/apptools/undo/traitsui/trait_set_command.py
@@ -1,0 +1,82 @@
+#
+# (C) Copyright 2015 Enthought, Inc., Austin, TX
+# All right reserved.
+#
+# This file is open source software distributed according to the terms in
+# LICENSE.txt
+#
+"""
+TraitSetCommand
+===============
+
+This module provides a simple command which sets a trait value on ``do`` and
+``redo``, resets it to the previous value on ``undo``, and merges multiple
+commands on the same attribute on the same object.
+
+"""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+# Enthought library imports
+from traits.api import Any, Str
+
+# Local library imports
+from apptools.undo.api import AbstractCommand
+
+
+class TraitSetCommand(AbstractCommand):
+    """ Command subclass which simply sets a trait on an object
+
+    This is somewhat simplistic: it assumes that nothing else is going to
+    hold on to and mutate the old value unexpectedly, which may break down
+    with things like NumPy arrays.  The main purposes is to generate commands
+    from the UndoHandler, so we can be reasonably comfortable that the values
+    are being handled sanely.
+
+    """
+
+    #-------------------------------------------------------------------------
+    # 'TraitSetCommand' interface
+    #-------------------------------------------------------------------------
+
+    #: The name of the data trait we are acting on.
+    trait_name = Str
+
+    #: The new value that the trait is being set to
+    value = Any
+
+    #: The old value of the trait
+    _old_value = Any
+
+    #-------------------------------------------------------------------------
+    # 'ICommand' interface
+    #-------------------------------------------------------------------------
+
+    #: The user-visible name of the command
+    name = Str
+
+    def do(self):
+        # we assume that copy or deep copy isn't required
+        self._old_value = getattr(self.data, self.trait_name)
+        self.redo()
+
+    def redo(self):
+        setattr(self.data, self.trait_name, self.value)
+
+    def undo(self):
+        setattr(self.data, self.trait_name, self._old_value)
+
+    def merge(self, other):
+        if (isinstance(other, self.__class__) and self.data is other.data and
+                self.trait_name == other.trait_name):
+            self.value = other.value
+            return True
+        return super(TraitSetCommand, self).merge(other)
+
+    # Traits default handlers
+
+    def _name_default(self):
+        label = self.data.trait(self.trait_name).label
+        if not label:
+            label = self.trait_name.replace('_', ' ')
+        return "Change "+label.title()
--- a/apptools/undo/traitsui/trait_set_command.py
+++ b/apptools/undo/traitsui/trait_set_command.py
@@ -18,7 +18,7 @@ commands on the same attribute on the same object.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 # Enthought library imports
-from traits.api import Any, Str
+from traits.api import Any, Bool, Str
 
 # Local library imports
 from apptools.undo.api import AbstractCommand
@@ -41,6 +41,9 @@ class TraitSetCommand(AbstractCommand):
 
     #: The name of the data trait we are acting on.
     trait_name = Str
+
+    #: whether additional commands can be merged together.
+    mergeable = Bool
 
     #: The new value that the trait is being set to
     value = Any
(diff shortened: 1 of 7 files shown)


tonysyu:
This reminds me, I had some undo code I wrote that never went into production. It sounds like the motivation here is similar in intent, at least: https://github.com/enthought/apptools/blob/feature/undo-helpers/apptools/undo/helpers/tests/test_base_undoable_traits_object.py

Some tests would be help clarify how this is used.

corranwebster:
Yes, the motivation is similar, but the implementation is quite different.  I'd be interested in seeing your code cleaned up a little and turned into a PR, as it seems like a useful complementary approach when you need finer-grained control over creating Commands from traits.  I wrote something similar for another project recently, although much more ad-hoc for that project's needs.

corranwebster:
Tests added.  They don't really help with showing how to use the code, though.  Perhaps an example or integration test might help.

jonathanrocher:
Looks simple and clean. An example would be useful indeed.

corranwebster:
I think that this is more-or-less complete.  Any blockers for a merge?

jonathanrocher:
@rkern Would you mind taking a look at this?

apptools/undo/traitsui/trait_set_command.py
```diff
@@ -0,0 +1,82 @@
...
+        setattr(self.data, self.trait_name, self.value)
+
+    def undo(self):
+        setattr(self.data, self.trait_name, self._old_value)
+
+    def merge(self, other):
+        if (isinstance(other, self.__class__) and self.data is other.data and
+                self.trait_name == other.trait_name):
+            self.value = other.value
+            return True
```
rkern: I am skeptical of this. This means that if I change the same value twice in a row, "Undo" will undo both settings, not just the last one.

corranwebster: The reason for doing this was because traitsui tends to generate many ephemeral changes: eg. if you enter "1000" into a text field you get the trait set to "1", then "10" then "100" and finally "1000".  You get similar behaviour with sliders and spinners as the value is being adjusted.  If we don't merge these, then you end up with lots of irrelevant entries on the undo stack.

If you feel strongly about this, I could add a cool-down period, so that if the next change comes in some time interval after the previous change (maybe a second or so) then it is treated as a separate undo event?  But that would increase the complexity of this code somewhat.

apptools/undo/traitsui/trait_set_command.py
```diff
@@ -0,0 +1,82 @@
...
+    #-------------------------------------------------------------------------
+    # 'ICommand' interface
+    #-------------------------------------------------------------------------
+
+    #: The user-visible name of the command
+    name = Str
+
+    def do(self):
+        # we assume that copy or deep copy isn't required
+        self._old_value = getattr(self.data, self.trait_name)
```
rkern: This will fail ungracefully for `Event` traits (`Button` presses). While it isn't possible to automatically handle the undoability of such UI actions, we need to avoid making a `TraitSetCommand` in such cases.

apptools/undo/traitsui/trait_set_command.py
rkern: I know, but this just indicates that this is not the right place to shim in undo functionality. It needs to be closer to the UI control since only some UI controls have this issue (admittedly, very important ones). The implementation of the `Editor` needs to be responsible for determining when estimating when a user has completed a single logical action even if it sends live trait updates throughout. E.g. a slider starts the logical action on button-down and completes it with button-up; the undo command should only be added to the stack on button-up. Similarly, a text field would "logically complete" for undo purposes on enter-pressed or focus-out (even if the `Editor` was configured to update the model live). Estimating logical completeness of the user's action based on the timing of the event stream is much less reliable, but might be workable (the tests will be a nightmare, of course). One option would be to allow trait or `Item` metadata that can inform the `UndoHandler` whether it should merge adjacent changes or not. So `TextEditors` and `RangeEditors` could merge while `EnumEditors` wouldn't, at the expense of making the coder be explicit about it.

Yes, the complexity of the code needs to increase and may need changes to the `Editors` themselves. It's a complex problem, and the information needed to handle it has been stripped out at this point of the framework. As it stands, I think this set of undocumented limitations makes this an attractive nuisance given where this code is located and how it is named. I think the user experience will be surprising in ways that diminish confidence in the user's perception of how safe it is to explore the UI, which is the point of building in undo functionality.

corranwebster: First question, then, is whether this whole approach of intercepting at the Handler level is worth pursuing further, in your opinion.  There are other approaches to solving this problem (such as Tony's branch above) which we could pursue for client projects, but they will require some effort, so I don't particularly want to take this further if you feel strongly that this is architecturally wrong.

I am already running into some cases where some more complex editors bypass the `Handler` completely when setting certain things (eg. selections) and so need separate shim code to capture the changes, but I think that is more a failing of the editor than an issue with using handlers, since such out-of-band changes defeat much of the point of custom handlers.

In any case, if you are of the opinion that the handler approach is worth pursing, I will implement something that allows editors, items and traits to express their mergeability, possibly with cool-down timers depending on my time constraints.  Although actual modifications to TraitsUI Editors and Items will have to wait - for the immediate needs of consulting project code trait metadata or ad-hoc Editor subclasses are probably enough.

apptools/undo/traitsui/trait_set_command.py
corranwebster: This has been handled in the `UndoHandler` by:

- detecting `event` traits and sending them through the standard `Handler` `setattr`
- trying to `push` the command, and if it fails (as it will for a write-only property, for example) with a `TraitError` trying to set it using the standard `Handler` `setattr` (the failure will mean that the command never gets added to the stack.

corranwebster:
@rkern I have a sketch implementation of control of merging using either:

- trait metadata
- editor attributes

No cool-down type implementation yet, and not sure I will try.  Also no attempt to find the `Item` associated with the operation.

Currently have it set so that subclasses of `TextEditor` and `RangeEditor` are mergeable by default unless the trait metadata says no.  In the future, concrete editors can grow an attribute or property `undo_mergeable` that will also be listened to by the command generator.

No tests right now, and there is some uncertainty about what to do if there are multiple editors for the same trait: there appears to be no way to tell which one generated the `setattr`.

Thoughts?

jonathanrocher:
The current state is broken and not passing its tests:
```
RocherMBP-7:tests jrocher$ nosetests -v
test_do (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_merge (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_merge_bad_command (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_merge_bad_data (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_merge_bad_trait (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_merge_not_mergeable (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_name_default (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_name_default_with_label (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_redo (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_undo (apptools.undo.traitsui.tests.test_trait_set_command.TestTraitSetCommand) ... ok
test_on_redo (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR
test_on_undo (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR
test_setattr (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR
test_setattr_bad (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR
test_setattr_event (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ok
test_setattr_property (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR
test_setattr_write_only (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler) ... ERROR

ERROR: test_on_redo (apptools.undo.traitsui.tests.test_undo_handler.TestUndoHandler)
  File "/Users/jrocher/Projects/ETS_trunk/apptools/apptools/undo/traitsui/tests/test_undo_handler.py", line 123, in test_on_redo
    self.handler.setattr(info, self.object, 'dummy_trait', 'new value')
  File "/Users/jrocher/Projects/ETS_trunk/apptools/apptools/undo/traitsui/undo_handler.py", line 67, in setattr
    mergeable = self._editor_mergeable(info, object, name)
  File "/Users/jrocher/Projects/ETS_trunk/apptools/apptools/undo/traitsui/undo_handler.py", line 91, in _editor_mergeable
    editors = info.ui.get_editors(name)
AttributeError: 'NoneType' object has no attribute 'ui'
```

corranwebster:
Closing.  Not the right way to do the integration.

PR #40, opened 2015-02-10
Reimplement LRUCache on top of OrderedDict.

johntyree:
This PR changes the newly implemented LRUCache to be based on `collections.OrderedDict` rather than the circular linked list ported from python 3. The linked list version had a bug wherein doubly-inserted values weren't properly handled and another involving recently dropped keys. It was sufficiently complicated that debugging was taking longer than reimplementing.

I'm much happier with this result anyway.

### DIFF
--- a/apptools/lru_cache/lru_cache.py
+++ b/apptools/lru_cache/lru_cache.py
@@ -25,8 +25,6 @@ class LRUCache(HasStrictTraits):
 
     Items older than `size()` accesses are dropped from the cache.
 
-    This is ported from python 3's functools.lru_cache.
-
     """
 
     size = Int
@@ -46,7 +44,7 @@ class LRUCache(HasStrictTraits):
         self._initialize_cache()
         super(LRUCache, self).__init__(**traits)
 
-    def _initialize_cache(self, force=False):
+    def _initialize_cache(self):
         with self._lock:
             if self._cache is None:
                 self._cache = OrderedDict()
@@ -54,8 +52,9 @@ class LRUCache(HasStrictTraits):
                 self._cache.clear()
 
     def _renew(self, key):
-        r = self._cache.pop(key)
-        self._cache[key] = r
+        with self._lock:
+            r = self._cache.pop(key)
+            self._cache[key] = r
         return r
 
     # -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


johntyree:
The py3 implementation had constant time operations while this almost surely doesn't, which is another facet to consider. I don't feel like the performance is anywhere near bad enough to worry about, however.

cfarrow:
Were those bugs in the original code or somehow introduced while porting?

apptools/lru_cache/lru_cache.py
```diff
@@ -10,21 +10,14 @@
...
-
-from traits.api import Callable, Dict, Event, HasStrictTraits, Instance, Int, \
-    List, Str
+from collections import OrderedDict
 
-logger = logging.getLogger(__name__)
+from traits.api import Callable, Event, HasStrictTraits, Instance, Int
 
 
 class LRUCache(HasStrictTraits):
```
cfarrow: The docstring refers to this as a port. I'm not sure that is still the case.

apptools/lru_cache/lru_cache.py
```diff
@@ -77,54 +71,17 @@ def __len__(self):
...
-                    # for last, after the root and links have been put in
-                    # a consistent state.
-                    self._cache[key] = oldroot
-                else:
-                    # Put result in a new link at the front of the queue.
-                    last = self._root[PREV]
-                    link = [last, self._root, key, result]
-                    last[NEXT] = self._root[PREV] = self._cache[key] = link
-            return None
+            self._cache[key] = result
```
cfarrow: Everything up to calling the callback should be under the lock. `OrderedDict` is not thread safe (http://bugs.python.org/issue11875#msg134089), so every addition and removal should be locked.

Beyond that, this contains a possible race condition:
1) Thread 1 just cached a value and execution just finished line 81.
2) Thread 2 just finished line 79. The key it is using is the current LRU.
3) Thread 1 executes line 82, popping the key and value just set by Thread 2.
4) Thread 2 eventually gets to line 80 and *crash*.

apptools/lru_cache/lru_cache.py
```diff
@@ -46,23 +39,24 @@ class LRUCache(HasStrictTraits):
...
-    _tag = Str("LRUCache")
+    _cache = Instance(OrderedDict)
 
     def __init__(self, size, **traits):
         self.size = size
         self._initialize_cache()
         super(LRUCache, self).__init__(**traits)
 
-    def _initialize_cache(self):
+    def _initialize_cache(self, force=False):
```
cfarrow: `force` is not used.

cfarrow:
I have a couple of comments, but the code otherwise looks good.

Performance is not an issue here. `OrderedDict` uses an internal linked list and node map to maintain constant time lookup and assignment.

apptools/lru_cache/lru_cache.py
johntyree: :+1:

johntyree:
The problem may have been introduced while porting because of the differences in the way we're using it here vs there, splitting up get/set etc. I haven't investigated, yet.

johntyree:
I don't think the bug is in the original. Just my sub-par adaptation :grimacing:

cfarrow:
More comments. (I left them on the latest commit, not the PR, sorry.)

cfarrow:
LGTM

Issue #35, opened 2014-10-03
apptools.preferences does not support storing unicode values.


sjagoe:
```python
Exception occurred in traits notification handler for object: <Preferences object at 0x40c19b0>, trait: last_name, old value: value, new value: TestUÜöser
Traceback (most recent call last):
  File "lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "lib/python2.7/site-packages/apptools/preferences/preferences_helper.py", line 71, in _anytrait_changed
    self.preferences.set('%s.%s' % (self._get_path(), trait_name), new)
  File "lib/python2.7/site-packages/apptools/preferences/scoped_preferences.py", line 166, in set
    node.set(path, value)
  File "lib/python2.7/site-packages/apptools/preferences/preferences.py", line 191, in set
    node.set('.'.join(components[1:]), value)
  File "lib/python2.7/site-packages/apptools/preferences/preferences.py", line 185, in set
    self._set(path, value)
  File "lib/python2.7/site-packages/apptools/preferences/preferences.py", line 546, in _set
    value = str(value)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 5-6: ordinal not in range(128)
```

dpinte:
Fixed by #41. Closing

PR #41, opened 2015-03-04
Fixing unicode support for PreferencesHelper

dpinte:
The PreferencesHelper don't work properly when getting/storing unicode strings with non-ascii caracters. This PR intends to fix the issue.

### DIFF
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -132,10 +132,9 @@ class PreferencesHelper(HasTraits):
         if isinstance(handler, Str) or trait.is_str:
             pass
 
-        # If the trait type is 'Unicode' then we convert the raw value.
         elif isinstance(handler, Unicode):
             # Just in case we get back an ASCII `str` object, convert it to a
-            # `unicode` object.
+            # `unicode` object.     
             value = unicode(value) 
 
         # Otherwise, we eval it!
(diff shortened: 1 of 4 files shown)


apptools/preferences/preferences_helper.py
```diff
@@ -134,7 +140,8 @@ def _get_value(self, trait_name, value):
 
         # If the trait type is 'Unicode' then we convert the raw value.
```
rkern: Man, that's very unreliable.

Can we just make `ConfigObj(file_or_filename, encoding='utf-8')` instead? What breaks when we do that?

dpinte: that's a great suggestion. I'll look into that as a better solution.

dpinte:
@rkern changing only ConfigObj is not enough as the Preferences class convert all the object to str internally. Either we need to change the way the preference sets the object by not converting them to str and letting ConfigObj doing the conversion job or we need to do the proper conversion from str to unicode with a decode/encode. What do you think? Changing the way the preference work by not converting the object to str is not hard. I am just slightly worried on the impact on external users.

dpinte:
@rkern 1f343e3 contains the other implementation. All the tests are passing with those changes. What do you think? 1f343e3 or 585c88b ?

dpinte:
@rkern I've updated the comments and the tests to do a full load/save cycle.

dpinte:
Fixes #35

rkern:
LGTM! Thanks!

PR #45, opened 2015-03-17
Use a full unicode layer in preferences.py

dpinte:
PR #41 was changing the behaviour of the Preference class by not converting object to str anymore. This caused a clear problem when getting preferences from a file as the same call to retrieve a preference could either return a specific type or a unicode object as read by ConfigObj in the file.  

This PR changes the behaviour again and gets back to the original design of Preferences which standardizes on str objects. When setting an object in the Preference class it is automatically converted to a unicode object (defaulting to utf-8 encoded strings as input), the unicode object is then properly serialized as utff-8 str by ConfigObj when saving the prefence to a file. When loading the preference in the preference helper, the object is nicely converted back to its target type, with `Str` as a special case where the unicode object is encoded to a utf-8 str)

This PR fixes the broken test suite.

@corranwebster Can you review? @rkern this changes a bit what we talked about in #41 but is probably the least painful implementation before switching to a better storage like yaml.

### DIFF
--- a/apptools/preferences/preferences.py
+++ b/apptools/preferences/preferences.py
@@ -541,6 +541,14 @@ class Preferences(HasTraits):
 
     def _set(self, key, value):
         """ Set the value of a preference in this node. """
+        
+        # everything must be unicode encoded so that ConfigObj configuration
+        # can properly serialize the data. If we receive regular str, they are
+        # supposed to be utf-8 encoded
+        if isinstance(value, str):
+            value = value.decode('utf-8')
+        else:
+            value = unicode(value)
 
         self._lk.acquire()
         old = self._preferences.get(key)
--- a/apptools/preferences/preferences.py
+++ b/apptools/preferences/preferences.py
@@ -543,12 +543,9 @@ class Preferences(HasTraits):
         """ Set the value of a preference in this node. """
         
         # everything must be unicode encoded so that ConfigObj configuration
-        # can properly serialize the data. If we receive regular str, they are
-        # supposed to be utf-8 encoded
-        if isinstance(value, str):
-            value = value.decode('utf-8')
-        else:
-            value = unicode(value)
+        # can properly serialize the data. Python str are supposed to be ASCII
+        # encoded.
+        value = unicode(value)
 
         self._lk.acquire()
         old = self._preferences.get(key)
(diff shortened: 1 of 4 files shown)


apptools/preferences/preferences.py
```diff
@@ -541,6 +541,14 @@ def _remove_preferences_listener(self, listener):
 
     def _set(self, key, value):
         """ Set the value of a preference in this node. """
+        
+        # everything must be unicode encoded so that ConfigObj configuration
+        # can properly serialize the data. If we receive regular str, they are
+        # supposed to be utf-8 encoded
+        if isinstance(value, str):
+            value = value.decode('utf-8')
```
rkern: I think that if we get a `str` from the Python side, we should assume it to be ASCII, not UTF-8. We should not be throwing around UTF-8-encoded `str` objects in our Python code.

apptools/preferences/preferences_helper.py
```diff
@@ -130,12 +130,13 @@ def _get_value(self, trait_name, value):
 
         # If the trait type is 'Str' then we just take the raw value.
         if isinstance(handler, Str) or trait.is_str:
-            pass
+            # Just in case we get back a unicode object, convert it to a
+            # utf-8 encoded object.
+            value = value.encode('utf-8')
```
rkern: `Str` traits are not `str` objects. They should be left as `unicode`.

apptools/preferences/preferences.py
dpinte: Thinking about my initial use case a bit more, what I was trying to solve was the unicode object coming from the UI (or other sources) and that had to be serialized to the preferences. I agree that the assumption of utf-8 encoded string is not good.. I'll remove that.

apptools/preferences/preferences_helper.py
dpinte: that's is even better then. I did not check that.

corranwebster:
@rkern's comments are appropriate, as always.  In addition, can you add this PR reference to this line in `CHANGES.txt`: https://github.com/enthought/apptools/blob/master/CHANGES.txt#L11

dpinte:
@rkern @corranwebster PR updated based on your comments. Does that look better?

corranwebster:
Actually, I meant #41 _and_ #45 in the `CHANGES.txt`, but I'll fix it to my satisfaction :)

corranwebster:
LGTM.

There is a difference in behaviour in that now a `str` value will round-trip and come back as a `unicode` value (but one which tests as equal to the original value), but I think that's OK.  If we don't want that behaviour, then that would add significant complexity to the code.

dpinte:
I agree. It is also simplified by the fact that string comparison between str and unicode is working.

dpinte:
thanks @corranwebster. Merging.

PR #52, opened 2015-04-24
Adding unit tests for CommandStack

jonathanrocher:
@corranwebster Feedback would be appreciated.


apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,119 @@
...
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+# -----------------------------------------------------------------------------
+
+from traits.testing.unittest_tools import unittest
+
+from apptools.undo.api import CommandStack, UndoManager
+from testing_commands import SimpleCommand, UnnamedCommand
```
tonysyu: Any particular reason for putting these in a separate module? It makes the tests harder to follow.

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def assert_n_commands_pushed(self, n):
+        self.assertEqual(len(self.stack._stack), n)
+        self.assertEqual(self.stack._index, n-1)
+        self.assertFalse(self.stack.clean)
+
+    def assert_n_commands_pushed_and_undone(self, n):
+        # N commands have been pushed
+        self.assertEqual(len(self.stack._stack), n)
+        self.assertEqual(self.stack._index, -1)
```
tonysyu: I'm sure there's a good reason for this, but it deserves a comment to clarify what's going on.

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,119 @@
...
+    # Assertion helpers -------------------------------------------------------
+
+    def assert_n_commands_pushed(self, n):
+        self.assertEqual(len(self.stack._stack), n)
+        self.assertEqual(self.stack._index, n-1)
+        self.assertFalse(self.stack.clean)
+
+    def assert_n_commands_pushed_and_undone(self, n):
+        # N commands have been pushed
+        self.assertEqual(len(self.stack._stack), n)
```
tonysyu: Tangential to this PR, `stack` should really define a `__len__` method.

apptools/undo/tests/test_command_stack.py
rkern: If they are in a separate file, they should not use implicit relative imports. `apptools.undo.tests` should be a package.

apptools/undo/tests/test_command_stack.py
tonysyu: Actually, I'm not convinced this `-1` makes sense. Would this work for `n > 1`?

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertEqual(self.stack._index, -1)
+        self.assertTrue(self.stack.clean)
+
+    # Tests -------------------------------------------------------------------
+
+    def test_empty_command_stack(self):
+        self.assertEqual(self.stack._stack, [])
+        self.assertEqual(self.stack._index, -1)
+        self.assertTrue(self.stack.clean)
```
tonysyu: Would `assert_n_commands_pushed(0)` work here (minus the check of `clean`, which is pretty mysterious based on the source)?

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,119 @@
...
+    def assert_n_commands_pushed(self, n):
+        self.assertEqual(len(self.stack._stack), n)
+        self.assertEqual(self.stack._index, n-1)
+        self.assertFalse(self.stack.clean)
+
+    def assert_n_commands_pushed_and_undone(self, n):
+        # N commands have been pushed
+        self.assertEqual(len(self.stack._stack), n)
+        self.assertEqual(self.stack._index, -1)
+        self.assertTrue(self.stack.clean)
```
tonysyu: Is this true for `n > 1`? `clean` is set to `True` when the stack is empty, but does it have anything to do with commands being undone? (I'm actually asking, b/c I don't know where the command-specific flag is being set.)

apptools/undo/tests/test_command_stack.py
jonathanrocher: This function is being used after pushing `n` commands and undoing them all. It is used below for `n=4`.

apptools/undo/tests/test_command_stack.py
jonathanrocher: Agreed.

apptools/undo/tests/test_command_stack.py
jonathanrocher: Good idea. I could/should use a test on `n` for the `clean=False` assertion.

apptools/undo/tests/test_command_stack.py
jonathanrocher: Thanks for the feedback. Moved to absolute import. Left in a different module so that it can be used for other unit tests on `Command` objects. @tonysyu I did it in a different file since the tests don't really need to know what the commands actually do. It only matters that they exists. Do you agree it doesn't prevent from following the tests?

apptools/undo/tests/test_command_stack.py
jonathanrocher: I added some comments to be more clear.

apptools/undo/tests/test_command_stack.py
jonathanrocher: Clarified offline. Code is correct. See new test added by 9bb6ddc.

tonysyu: When we were discussing this earlier, you clarified why the stack index is `-1`. I still think the `clean` state is tangential to the number of commands pushed and undone.

tonysyu:
After a bit more thought, I think these tests would be clearer if the asserts were `assert_n_undoable_commands` and `assert_n_redoable_commands`, rather than testing the number of commands pushed.

apptools/undo/tests/test_command_stack.py
jonathanrocher: I agree with you @tonysyu . I have removed the cleanliness tests from the assertions and followed your suggestion by making them context managers.

apptools/undo/tests/test_command_stack.py
tonysyu: I'm not a fan of turning `tests` directories into packages or separating test data from tests, but if you prefer it, then that's fine.

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,143 @@
...
+    # Assertion helpers -------------------------------------------------------
+from contextlib import contextmanager
+from nose.tools import assert_equal
+
+@contextmanager
+def assert_n_commands_pushed(stack, n):
+    current_length = len(stack._stack)
+    try:
+        yield
+    finally:
```
tonysyu: I think this is one of the cases where a `try/finally` block isn't appropriate for a context manager. I think this will suppress errors in the test and, instead, raise assertion errors.

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,143 @@
...
+        # ... and the state is at the tip of the stack...
+        assert_equal(stack._index, current_length+n-1)
+
+
+@contextmanager
+def assert_n_commands_pushed_and_undone(stack, n):
+    current_length = len(stack._stack)
+    try:
+        yield
+    finally:
```
tonysyu: Same as above: remove `try/finally`

tonysyu:
Just a couple of comments about `try/finally` blocks; otherwise, LGTM.

apptools/undo/tests/test_command_stack.py
```diff
@@ -0,0 +1,143 @@
...
+    try:
+        for i in range(num_commands):
+            stack.push(command)
+    finally:
+        stack.end_macro()
+
+
+    # Assertion helpers -------------------------------------------------------
+from contextlib import contextmanager
+from nose.tools import assert_equal
```
jwiggins: Imports go at the top :wink:

jonathanrocher: Thanks :). Fixed.

jonathanrocher:
All comments addressed. I pushed a couple more tests around cleanliness. Final blessing from @corranwebster since this is your code or @jwiggins since you are still officially the maintainer?

jwiggins:
:+1:

jonathanrocher:
Merging then. Thanks all for feedback.

PR #61, opened 2016-03-04
Fix state_pickler.dump in py2

kitchoi:
In Python2 environment, `state_pickler.dump(obj, u'file.path')` would raise an Error, but it is not necessary to do further type checking as `open` already handles that for us.

@prabhuramachandran can you take a look please?  This problem arises when trying Save/Load visualization in Mayavi2 in Python2.7 environment.

### DIFF
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -918,18 +918,14 @@ class StateSetter:
 def _get_file_read(f):
     if hasattr(f, 'read'):
         return f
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
@kitchoi -- thanks!  Do all the mayavi tests pass with this change?

kitchoi:
@prabhuramachandran Yes tests for mayavi pass.

prabhuramachandran:
Thanks, merging!

PR #62, opened 2016-04-08
Handle numpy pure scalar types

stefanoborini:
spawned from https://github.com/enthought/mayavi/issues/61
The state pickler/unpickler could not handle numpy scalar types,
which were forced to None. This meant that trait classes having
Float() trait hosting a numpy.float64 could not be pickled
(to be precise they were pickled, with the content being None, silently).

### DIFF
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -480,6 +480,7 @@ class StatePickler:
 
     def _do_numpy_generic_type(self, value):
         idx = self._register(value)
+        self._misc_cache.append(value)
         data = base64_encode(pickle.dumps(value))
         return dict(type='numpy', id=idx, data=data)
 
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -481,7 +481,7 @@ class StatePickler:
     def _do_numpy_generic_type(self, value):
         idx = self._register(value)
         self._misc_cache.append(value)
-        data = base64_encode(pickle.dumps(value))
+        data = base64_encode(value.dumps())
         return dict(type='numpy', id=idx, data=data)
 
 
@@ -700,7 +700,7 @@ class StateUnpickler:
         return result
 
     def _do_numpy_generic_type(self, value, path):
-        result = pickle.loads(base64_decode(value["data"]))
+        result = numpy.loads(base64_decode(value["data"]))
         self._obj_cache[value['id']] = result
         return result
 
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -142,7 +142,7 @@ def gunzip_string(data):
 
 def base64_encode(data):
     if PY_VER > 2:
-        base64.encodebytes(data)
+        return base64.encodebytes(data)
     else:
         return base64.encodestring(data)
 
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -100,10 +100,10 @@ Notes
 # Standard library imports.
 import base64
 import sys
-import types
 import pickle
 import gzip
 from io import BytesIO, StringIO
+import warnings
 
 import numpy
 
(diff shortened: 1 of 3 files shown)


stefanoborini:
Failures were already in master and are due to 2.6 and 3.4 incompatibilities. 
@dpinte do we still care about 2.6 support? I might have to fix master for 3.4 and merge back onto this to get green.

dpinte:
@stefanoborini I think we can start dropping the 2.6 support in apptools

stefanoborini:
@dpinte ok.

stefanoborini:
Failure is related to 3.4 and unrelated to the change. I am going to bisect master and see where it was introduced.

stefanoborini:
Strangely enough, tests pass on my machine with python 3.4.

apptools/persistence/state_pickler.py
```diff
@@ -341,8 +360,19 @@ def _do(self, obj):
...
+                          " for backward compatibility.".format(obj_type))
+            return None
 
     def _get_id(self, value):
+        # We consider a special case for numpy scalar values, because
+        # they hash as native types, but they are special because we
+        # want to recover their true type, and the only way of doing so
+        # is to consider them as objects.
+        if isinstance(value, numpy.generic):
+            return id(value)
```
itziakos: This is a small problem with using `id` here. The same `id` can be re-used when the original object is garbaged collected. So if during the pickling process the numpy arrays are garbage collected then the registry might point to the wrong object.

itziakos: I have updated the pickling do method to keep the object alive in the cash (for the duration of the pickling process). This should be enough to avoid the `id` reuse case.

apptools/persistence/state_pickler.py
```diff
@@ -445,12 +475,14 @@ def _do_dict(self, value):
...
-            data = base64.encodebytes(gzip_string(numpy.ndarray.dumps(value)))
-        else:
-            data = base64.encodestring(gzip_string(numpy.ndarray.dumps(value)))
+        data = base64_encode(gzip_string(numpy.ndarray.dumps(value)))
         return dict(type='numeric', id=idx, data=data)
 
+    def _do_numpy_generic_type(self, value):
+        idx = self._register(value)
+        self._misc_cache.append(value)
+        data = base64_encode(pickle.dumps(value))
```
itziakos: Pickling numpy scalars is quite costly (in space). I wonder if there is a better way.

rkern: Use the `HIGHEST_PROTOCOL`

rkern: But otherwise, not so much. You'll end up reproducing exactly what `pickle` does anyways, if you want to be general.

rkern: `data = base64_encode(value.dumps())`

itziakos:
@prabhuramachandran can you have a look. is it ok to merge this PR?

stefanoborini:
Imho, it's good for me. Good spot.

apptools/persistence/state_pickler.py
```diff
@@ -341,8 +360,19 @@ def _do(self, obj):
...
     def _get_id(self, value):
+        # We consider a special case for numpy scalar values, because
+        # they hash as native types, but they are special because we
+        # want to recover their true type, and the only way of doing so
+        # is to consider them as objects.
+        if isinstance(value, numpy.generic):
+            return id(value)
+
         try:
             key = hash(value)
```
mdickinson: The use of `hash(value)` as a key is broken: it's assuming that hash will be one-to-one. This is orthogonal to the issue fixed by this PR; I'll open a separate issue.

apptools/persistence/state_pickler.py
mdickinson: Is there a simple way to avoid the caching altogether? The current caching code is broken (#66), and I don't think it makes a lot of sense to be caching simple immutable scalars in the first place: the size of the "reference" dictionary generated is going to be comparable to the size of the object dictionary.

apptools/persistence/state_pickler.py
mdickinson: > I'll open a separate issue.

#66

PR #64, opened 2016-06-13
NotViewer: the Traits Notification Viewer

pberkes:
A basic TraitsUI viewer for traits notifications. It's easy to include it in an existing application to monitor and debug traits notifications. See the docs for more information!

This is the same as enthought/etsdevtools#2 , in `apptools`, with docs.


kitchoi:
@pberkes (👋 ), this does look useful - thank you for this!

From offline discussion: The `traits.util.event_tracer` required by this application might be reworked, for example, to allow multiple tracers to be attached. It would be harder to do that refactoring if this application had been made publicly available first. This is still valuable code to keep. The last commit has been tagged as `experimental/notification-viewer` so that it can be easily recovered later.

Closing for now. But thanks a lot!

PR #73, opened 2017-05-31
Fix mismatched PyFace and Traits UI requirements.

mdickinson:
We're using Traits UI master, but only using PyFace master for Python 3. The PyFace and Traits UI requirements need to be matched on both Python 2 and Python 3.


corranwebster:
LGTM

Issue #71, opened 2017-01-12
Consider moving Workbench from Pyface to Apptools


corranwebster:
See https://github.com/enthought/pyface/issues/249  This would impact Mayavi and Envisage, but only as far as changing import locations.

@rkern is this something we should be considering?

rkern:
That's not unreasonable.

PR #75, opened 2018-01-03
[WIP] Move Workbench from Pyface

corranwebster:
See https://github.com/enthought/pyface/issues/249 and https://github.com/enthought/apptools/issues/71.

This is currently just taking the code from Pyface, renaming directories, and updating imports.  An attempt has been made to keep the commit history of the code.

The plan would be to merge this soon into apptools master, and then following the next apptools release have dependent projects switch their imports to point to apptools rather than Pyface.  In the longer term (currently scheduled for Pyface 7.0.0), Workbench would be removed completely from Pyface.

To do:
- [ ] update Apptools CI to perform tests with PyQt and Wx backends.


mdickinson:
@corranwebster What's the status of this PR?

corranwebster:
This is out of date, and we will probably make workbench a stand-alone package.  Closing.

PR #82, opened 2019-01-23
FIX: Use trait_set instead of set

larsoner:
Fixes warnings of the form:
```
apptools/preferences/preferences_helper.py:166: DeprecationWarning: use "HasTraits.trait_set" instead
    self.set(trait_change_notify=notify, **traits_to_set)
```
Let me know if this needs to be a `try: self.trait_set(...); except AttributeError: self.set(...)` for backward compat and I can change it.

### DIFF
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -163,7 +163,7 @@ class PreferencesHelper(HasTraits):
                 value = self._get_value(trait_name, preferences.get(key))
                 traits_to_set[trait_name] = value
 
-        self.set(trait_change_notify=notify, **traits_to_set)
+        self.trait_set(trait_change_notify=notify, **traits_to_set)
 
         # Listen for changes to the node's preferences.
         preferences.add_preferences_listener(


mdickinson:
Thanks for this. A quick `"grin "\(.set\("` shows me a couple other suspicious uses of `.set` besides this one (though they're obviously not being exercised by the test suite. Grr):
```
./apptools/preferences/preference_binding.py:
  142 :             self.obj.set(trait_change_notify=notify, **traits)
./apptools/preferences/ui/widget_editor.py:
   80 :         return self.set(**traits)
./apptools/template/test/scatter_plot.py:
  195 :         plot.set(  padding_left   = 50,
```

Are you interested in fixing any of these in this PR? If so, I'll wait, but if not, this looks good to merge.

> Let me know if this needs to be a `try: self.trait_set(...); except AttributeError: self.set(...)` for backward compat and I can change it.

No, no need; I haven't checked what ancient version of Traits would need `set` rather than `trait_set`, but for sure it's one we don't care about any more. We've been consistently replacing `.set` with `.trait_set` in other libraries.

larsoner:
@mdickinson done

PR #83, opened 2019-01-23
FIX: Fix for NumPy deprecation

larsoner:
Following the suggestion from:
```
  /home/larsoner/python/apptools/build/lib/apptools/persistence/state_pickler.py:661: DeprecationWarning: np.loads is deprecated, use pickle.loads instead
    result = numpy.loads(junk, encoding='bytes')
```


mdickinson:
LGTM. Thank you!

mdickinson:
FWIW, here's the current source of `numpy.loads`, making it clear that `pickle.loads` is a drop-in replacement: https://github.com/numpy/numpy/blob/a5cace49b79690795df29c6da3587e7a45a41591/numpy/lib/npyio.py#L38-L44

PR #85, opened 2019-02-03
Fixing auto-build problems with small containers

stewbond:
Self tests will fail if we build this package in an isolated chroot environment. 

That's because os.removedirs() was used.  This deletes parent directories if empty.  In our case, we were building with root permissions (normal for chroot), in an isolated enviornment.  The /tmp directory was empty, so when tests ran os.removedirs(), /tmp was also deleted.  The next test that tried to tempfile.mkdtemp() would fail because it couldn't create a new directory in /tmp because /tmp didn't exist.

### DIFF
--- a/apptools/preferences/tests/py_config_file_test_case.py
+++ b/apptools/preferences/tests/py_config_file_test_case.py
@@ -122,7 +122,7 @@ class PyConfigFileTestCase(unittest.TestCase):
         finally:
             # Clean up!
             os.remove(tmp)
-            os.removedirs(tmpdir)
+            os.rmdirs(tmpdir)
 
         return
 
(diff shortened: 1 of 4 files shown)


rahulporuri:
closing and opening to retrigger CI

apptools/preferences/tests/py_config_file_test_case.py
```diff
@@ -122,7 +122,7 @@ def test_save(self):
         finally:
             # Clean up!
             os.remove(tmp)
-            os.removedirs(tmpdir)
+            os.rmdirs(tmpdir)
```
rahulporuri: Spelling mistake. This should be `os.rmdir`, not `os.rmdirs`.

mdickinson:
```suggestion
            os.rmdir(tmpdir)
```

rahulporuri:
closing in favor of #178 . I merged master into this branch and fixed the typo.

PR #88, opened 2019-02-03
Skipping a unit test when run as root

stewbond:
When building and testing the package in a chroot enviornment, we have a root user.  However, test_properties fails when run as a root user.

```
os.chmod(f.path, stat.S_IRUSR)
self.assertEqual(f.is_readonly, True)
```

This fails because root will always have permission to write to the file.  That makes sense.  And it also therefore makes sense to skip this test when run as root.

### DIFF
--- a/apptools/io/tests/file_test_case.py
+++ b/apptools/io/tests/file_test_case.py
@@ -51,7 +51,7 @@ class FileTestCase(unittest.TestCase):
     ###########################################################################
     # Tests.
     ###########################################################################
-
+    @unittest.skipIf(os.getuid() == 0, "Cannot be run as root")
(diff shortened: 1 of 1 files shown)


mdickinson:
I'm not convinced that this is the right solution. For one thing, `os.getuid` doesn't exist on Windows.

mdickinson:
Closing and re-opening to retrigger CI.

kitchoi:
Thank you @stewbond. The issue looks like a legitimate one. I opened #191 to track it.

rahulporuri:
closing as we now have an issue to track the underlying problem and this solution doesn't work on all platforms so it's not the right one for us.

PR #90, opened 2019-03-12
Remove use of 2to3

rahulporuri:
Use `six` instead

~WARNING : I used `modernize` to make automated changes to the codebase. I then added the recommended changes without much review. As such, this PR might include changes which are not absolutely necessary to the functionality of the package. If necessary, I can review the changes and remove the unnecessary/overzealous changes.~


rahulporuri:
The failures are now Python 2 only. See job https://travis-ci.org/enthought/apptools/jobs/531612677

jwiggins (commented):
I think the blanket use of `from __future__ import absolute_import` is of dubious usefulness.

Additionally, it might be worth the time to examine each of the files where `six.moves` is being used. Specifically, `map`, `zip`, and `range` might be OK as is.

apptools/template/impl/any_context_data_name_item.py
```diff
@@ -122,13 +124,13 @@ def inputs_changed ( self ):
             filter = self.filter
```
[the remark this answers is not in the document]
rahulporuri: looks like it. corrected.

apptools/persistence/state_pickler.py
```diff
@@ -440,7 +444,7 @@ def _do_list(self, value):
     def _do_dict(self, value):
         idx = self._register(value)
         vals = [self._do(x) for x in value.values()]
-        data = dict(zip(value.keys(), vals))
+        data = dict(list(zip(list(value.keys()), vals)))
```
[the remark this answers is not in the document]
rahulporuri: neither of them are. corrected.

apptools/persistence/state_pickler.py
```diff
@@ -270,7 +274,7 @@ def __init__(self):
                     }
         if PY_VER == 2:
             type_map[long] = self._do_basic_type
-            type_map[unicode] = self._do_basic_type
+            type_map[six.text_type] = self._do_basic_type
```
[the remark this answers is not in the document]
rahulporuri: yes but this should not have been touched as it is python 2-only code.

```
>>> import six
>>> six.text_type
<type 'unicode'>
```

apptools/appscripting/script_manager.py
```diff
@@ -23,11 +24,12 @@
...
-from i_bind_event import IBindEvent
-from i_script_manager import IScriptManager
-from lazy_namespace import add_to_namespace, FactoryWrapper, LazyNamespace
-from scriptable_type import make_object_scriptable
+from .bind_event import BindEvent
+from .i_bind_event import IBindEvent
+from .i_script_manager import IScriptManager
+from .lazy_namespace import add_to_namespace, FactoryWrapper, LazyNamespace
+from .scriptable_type import make_object_scriptable
+import six
```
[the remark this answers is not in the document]
corranwebster: Unfortunately modernize doesn't understand import order conventions :(

apptools/io/__init__.py
```diff
@@ -14,4 +14,4 @@
 """ Provides an abstraction for files and folders in a file system.
     Part of the AppTools project of the Enthought Tool Suite.
 """
-from api import *
+from .api import *
```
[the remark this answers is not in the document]
corranwebster: We could at least make it explicit imports.

apptools/lru_cache/tests/test_lru_cache.py
```diff
@@ -141,7 +144,7 @@ def test_updated_event():
     c.on_trait_change(lambda x: events.append(x), 'updated')
 
     c[0] = 0
-    print(c.keys())
+    print(list(c.keys()))
```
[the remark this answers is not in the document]
rahulporuri: removing. maybe it was a left over debugging statement.

apptools/undo/tests/test_command_stack.py
```diff
@@ -12,14 +12,15 @@
...
 
-from __future__ import absolute_import
-from traits.testing.unittest_tools import unittest
 from contextlib import contextmanager
+
 from nose.tools import assert_equal
+from six.moves import range
 
+from traits.testing.unittest_tools import unittest
+from traits.testing.unittest_tools import unittest
```
jwiggins: Duped.

apptools/io/h5/dict_node.py
```diff
@@ -69,7 +69,7 @@ def __contains__(self, key):
         return key in self.data
 
     def keys(self):
-        return self.data.keys()
+        return list(self.data.keys())
```
rahulporuri: This change needs discussion.

If we don't make this change, we have different behavior on Python 2 and Python 3.
If wrap the call with `list`, the return value of `H5DictNode.keys` differs from how a normal dictionary behaves.

apptools/lru_cache/lru_cache.py
```diff
@@ -98,15 +98,15 @@ def get(self, key, default=None):
 
     def items(self):
         with self._lock:
-            return self._cache.items()
+            return list(self._cache.items())
```
rahulporuri: Similar to an earlier question regd H5DictNode, should the items/keys/values methods return iterables on Python 3 or should they return lists?

apptools/io/__init__.py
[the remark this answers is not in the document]
rahulporuri: Made them full imports ("from apptools.io.api import *")

apptools/persistence/tests/test_state_pickler.py
```diff
@@ -5,17 +5,24 @@
...
 import numpy
+import six
 
 from traits.api import Bool, Int, Long, Array, Float, Complex, Any, \
     Str, Unicode, Instance, Tuple, List, Dict, HasTraits
 
+if six.PY2:
+    long_trait_type = Long
+else:
+    long_trait_type = Int
```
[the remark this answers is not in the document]
rahulporuri: Hmm. I checked and the Int and Long don't share a common baseclass on Python 3. But they do use the same/similar validator for traits on Python 3.

Either way, this code isn't necessary. It was added when I was debugging an issue that this PR added! Removed these changes.

apptools/io/__init__.py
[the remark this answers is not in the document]
jwiggins: I'm a little surprised that that doesn't cause a recursion error...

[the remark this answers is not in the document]
jwiggins: (also, I was more horrified by the `*` than the relative import)

apptools/io/h5/table_node.py
```diff
@@ -1,6 +1,9 @@
+from __future__ import absolute_import
 import numpy as np
 from pandas import DataFrame
 from tables.table import Table as PyTablesTable
+import six
+from six.moves import zip
```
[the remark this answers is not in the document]
rahulporuri: The builtin `zip` is adequate. I removed this.

apptools/io/h5/dict_node.py
jwiggins: I think that because this method is part of the dictionary interface then it should act like a normal dictionary. Remove the `list` wrapper.

apptools/lru_cache/lru_cache.py
jwiggins: I want to say they should return iterables, but the locking presents a thorny problem. Perhaps the safest solution would be to implement this as an iterator (explicitly `yield`-ing items) on Python 3, and using the old/existing implementation on Python 2.

rahulporuri: I'd prefer leaving this change as-is, mainly because we're interested in removing `LRUCache`. Is that okay?

rahulporuri:
@jwiggins @corranwebster this PR is ready for review. again. for the last time, i thnk.

apptools/persistence/spickle.py
```diff
@@ -15,6 +15,7 @@
 # Copyright (c) 2006-2015, Prabhu Ramachandran
 # License: BSD Style.
 
+from __future__ import absolute_import
 import sys
 if sys.version_info[0] > 2:
     raise ImportError("This module does not work with Python 3.x")
```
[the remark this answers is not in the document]
rahulporuri: I tried using the `exclude` kwarg when calling `find_packages` to exclude just this module from the Python 3 package but it looks like `exclude` will only help when entire packages/sub-packages need to be excluded and not when one specific module needs to be excluded.

apptools/io/h5/table_node.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 from pandas import DataFrame
 from tables.table import Table as PyTablesTable
+import six
```
jwiggins: Move this up so that it's after the pandas import

apptools/io/h5/tests/test_file.py
```diff
@@ -509,15 +509,15 @@ def test_attribute_iteration_methods():
...
         attrs['alpha'] = 0xff
 
-        items = attrs.items()
+        items = list(attrs.items())
 
         assert all(isinstance(x, tuple) for x in items)
 
         # unfold the pairs
-        keys, vals = map(list, zip(*items))
+        keys, vals = [list(item) for item in zip(*items)]
```
jwiggins: No idea why this was originally written in such a complicated way. Just write `keys, vals = zip(*items)`. It's a common Python idiom and everything.

apptools/lru_cache/tests/test_lru_cache.py
```diff
@@ -73,7 +73,7 @@ def test_cache_resize():
 
 def test_cache_items():
     c = LRUCache(2)
-    assert_equal([], c.items())
```
jwiggins: `LRUCache.items` returns a `list` already.

apptools/lru_cache/tests/test_lru_cache.py
```diff
@@ -105,15 +105,15 @@ def test_cache_idempotency():
 
 def test_cache_keys_values():
     c = LRUCache(2)
-    assert_equal([], c.items())
+    assert_equal([], list(c.items()))
```
jwiggins: `LRUCache.items` returns a `list` already.

apptools/naming/trait_defs/naming_traits.py
```diff
@@ -27,6 +27,7 @@
 
 from apptools.naming.api \
     import Binding
+import six
```
jwiggins: This should probably be moved to above the ETS imports

apptools/naming/ui/context_node_type.py
```diff
@@ -15,11 +15,13 @@
 
 
 # Enthought library imports.
+from __future__ import absolute_import
+from __future__ import print_function
```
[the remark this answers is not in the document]
jwiggins: Also the `# Enthought library imports.` comment above this is a bit off

apptools/preferences/preferences.py
```diff
@@ -1,8 +1,8 @@
 """ The default implementation of a node in a preferences hierarchy. """
 
-from __future__ import print_function
-
 # Standard library imports.
+from __future__ import absolute_import
+from __future__ import print_function
```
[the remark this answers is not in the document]
jwiggins: Useless diff now

apptools/preferences/tests/preferences_test_case.py
```diff
@@ -164,14 +164,14 @@ def test_keys(self):
...
+        self.assertEqual([], list(p.keys()))
 
         # Set some preferences in the node.
         p.set('a', '1')
         p.set('b', '2')
         p.set('c', '3')
 
-        keys = p.keys()
+        keys = list(p.keys())
         keys.sort()
```
jwiggins: One liner: `keys = sorted(p.keys())`

apptools/preferences/ui/preferences_node.py
```diff
@@ -15,11 +15,13 @@
 
 
 # Enthought library imports.
+from __future__ import absolute_import
+from __future__ import print_function
```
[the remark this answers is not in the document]
jwiggins: Not an Enthought library import

apptools/scripting/recorder.py
```diff
@@ -221,9 +221,9 @@ def register(self, object, parent=None, trait_name_on_parent='',
             # Always ignore these.
             ignore.extend(['trait_added', 'trait_modified'])
 
-            sub_recordables = object.traits(record=True).keys()
+            sub_recordables = list(object.traits(record=True).keys())
             # Find all the trait names we must ignore.
-            ignore.extend(object.traits(record=False).keys())
+            ignore.extend(list(object.traits(record=False).keys()))
```
jwiggins: You can pass an iterable to `list.extend`

apptools/sweet_pickle/tests/updater_test_case.py
```diff
@@ -238,8 +238,8 @@ def fn2():
         self.assertEqual(2, len(self.registry.state_functions))
         self.assertEqual(2, len(self.registry.version_attribute_map))
         self.assertEqual(1, len(self.registry._state_function_classes))
-        self._validate_class_map_contents(updater.class_map.items() + \
-            updater2.class_map.items())
+        self._validate_class_map_contents(list(updater.class_map.items()) + \
```
jwiggins: The backtick is unnecessary

apptools/io/h5/tests/test_file.py
mdickinson: FWIW, I'm really not a fan of the `zip(*stuff)` idiom: it's got a nasty not-immediately-obvious corner case where `stuff` is empty. Fine in this case, but then it gets re-used in places where it's not fine and people don't notice the corner case until it's too late.

```
>>> d = {'a': 1, 'b': 2}
>>> keys, values = zip(*d.items())  # fine
>>> d = {}
>>> keys, values = zip(*d.items())  # not fine
```

mdickinson: This doesn't detract from @jwiggins' comment, of course.

One possible reason for the `list` is that otherwise `keys` and `values` are tuples. I haven't looked at the context, so I don't know if that matters in this particular case, but that's the usual reason for the extra `map(list, ...)` construct.

rahulporuri: `keys`, `vals` end up as tuples when we just use `= zip(*items)`, which is not expected as we then try comparing to lists. I'm keeping the `[list(item) ... ]` code instead of updating other code to test against tuples or wrap `keys`/`vals` with `list` later in the code.

jwiggins: Now that you've reminded me that `zip` returns tuples, that explains why it was initially written as `map(list, zip(*items))`.

jwiggins (approved):

PR #92, opened 2019-05-18
Use etstool utility

rahulporuri:
~NOTE : The base branch for this PR is #90~

The `etstool` utility module is used to setup the environments and test the package. Travis was updated to use the utility module when installing and testing the package. Note that we are now testing the package on OSX along with linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,6 @@ install:
   - edm run -- python etstool.py install --runtime=${RUNTIME} || exit
 script:
   - edm run -- python etstool.py test --runtime=${RUNTIME} || exit
-  - edm run -- python etstool.py docs --runtime=${RUNTIME} || exit
 
 after_success:
   - edm run -- coverage combine
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ cache:
 
 before_install:
   - mkdir -p "${HOME}/.cache/download"
-  - if [[ ${TRAVIS_OS_NAME} == "linux" ]] ; then sh -e /etc/init.d/xvfb start; fi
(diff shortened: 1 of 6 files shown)


rahulporuri:
The travis CI failure is 
```
FAIL: Test if relative paths are set correctly.
  File "/home/travis/build/enthought/apptools/apptools/persistence/tests/test_file_path.py", line 48, in test_relative
    self.assertEqual(f.rel_pth, expect)
AssertionError: 'tmpAO5BM4/data/t.vtk' != '../tmp/tmpAO5BM4/data/t.vtk'
```

rahulporuri:
i'm not sure why we don't see this failure in #90

rahulporuri:
What's more interesting is the fact that this failure is only seen on the linux travis jobs, not on the osx travis jobs. I feel like i'm missing something obvious here.

rahulporuri:
sigh. i'm not sure what's going on with the CI. all of the linux jobs failed with the following error - 

```

FAIL: Test if relative paths are set correctly.

  File "/home/travis/build/enthought/apptools/apptools/persistence/tests/test_file_path.py", line 48, in test_relative

    self.assertEqual(f.rel_pth, expect)

AssertionError: 'tmpaEZMUG/data/t.vtk' != '../tmp/tmpaEZMUG/data/t.vtk'

```

which is also the error i saw in #92

mdickinson:
The test failure is a direct consequence of running the tests from a directory that's under `/tmp`.

We can fix the test, or we can run the tests from a newly-created empty directory in the repository instead of running from a temporary directory. The former would be better (obviously).

mdickinson (approved):
LGTM. The test failure should (?) be fixed by merging in #104.

PR #95, opened 2019-07-03
Update cStringIO imports

### LINKED ISSUE #94: No module named 'cStringIO' under Python 3
apptools for Python 3 may not have module named 'cStringIO', please have a look at: https://stackoverflow.com/q/28200366

Related issue: enthought/mayavi/issues/797

rahulporuri:
fixes #94

### DIFF
--- a/apptools/logger/agent/attachments.py
+++ b/apptools/logger/agent/attachments.py
@@ -8,8 +8,8 @@ rewrite.
 
 import logging
 import os.path
-from email import Encoders
-from email.MIMEBase import MIMEBase
+from email import encoders
+from email.mime.base import MIMEBase
 
 from traits.api import Any, HasTraits
 
@@ -71,7 +71,7 @@ class Attachments(HasTraits):
 
         msg.set_payload(file_object.getvalue())
 
-        Encoders.encode_base64(msg) # Encode the payload using Base64
+        encoders.encode_base64(msg) # Encode the payload using Base64
         msg.add_header('Content-Disposition', 'attachment', filename='project.zip')
 
         self.message.attach(msg)
--- a/apptools/logger/agent/attachments.py
+++ b/apptools/logger/agent/attachments.py
@@ -58,7 +58,7 @@ class Attachments(HasTraits):
         relpath = os.path.basename(dir)
 
         import zipfile
-        from cStringIO import StringIO
+        from six.moves import cStringIO as StringIO
 
         ctype = 'application/octet-stream'
         maintype, subtype = ctype.split('/', 1)
--- a/apptools/logger/agent/attachments.py
+++ b/apptools/logger/agent/attachments.py
@@ -58,13 +58,13 @@ class Attachments(HasTraits):
         relpath = os.path.basename(dir)
 
         import zipfile
-        from six.moves import cStringIO as StringIO
+        from six import BytesIO
 
         ctype = 'application/octet-stream'
         maintype, subtype = ctype.split('/', 1)
         msg = MIMEBase(maintype, subtype)
 
-        file_object = StringIO()
+        file_object = BytesIO()
         zip = zipfile.ZipFile(file_object, 'w')
         _append_to_zip_archive(zip, dir, relpath)
         zip.close()
--- a/apptools/logger/agent/attachments.py
+++ b/apptools/logger/agent/attachments.py
@@ -58,7 +58,7 @@ class Attachments(HasTraits):
         relpath = os.path.basename(dir)
 
         import zipfile
-        from six import BytesIO
+        from io import BytesIO
 
         ctype = 'application/octet-stream'
         maintype, subtype = ctype.split('/', 1)
--- a/apptools/logger/agent/tests/test_attachments.py
+++ b/apptools/logger/agent/tests/test_attachments.py
@@ -21,7 +21,7 @@ class AttachmentsTestCase(unittest.TestCase):
         self.addCleanup(shutil.rmtree, self.tmpdir)
 
         self.tmpfile = os.path.join(self.tmpdir, "dummy_file.txt")
-        with io.open(self.tmpfile, 'w') as filehandle:
+        with io.open(self.tmpfile, 'w', encoding='utf8') as filehandle:
             filehandle.write(u"Dummy data in dummy file for dummies")
 
     def test_attaching_workspace(self):
--- a/apptools/logger/agent/tests/test_attachments.py
+++ b/apptools/logger/agent/tests/test_attachments.py
@@ -42,7 +42,7 @@ class AttachmentsTestCase(unittest.TestCase):
         message = attachments.message
         self.assertTrue(message.is_multipart())
         payload = message.get_payload()
-        self.assertTrue(len(payload), 1)
+        self.assertEqual(len(payload), 1)
 
     def test_attaching_single_project(self):
         class DummySingleProject(object):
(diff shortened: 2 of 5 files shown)


apptools/logger/agent/attachments.py
```diff
@@ -58,7 +58,7 @@ def _attach_directory(self, dir):
         relpath = os.path.basename(dir)
 
         import zipfile
-        from cStringIO import StringIO
+        from six.moves import cStringIO as StringIO
```
mdickinson: Would the simpler `from six import StringIO` work here?

mdickinson: Actually, looking at the place where this is used, do we need `BytesIO` rather than `StringIO`?

mdickinson: Looks like we do. On Python 3, working in a directory that contains a `tox.ini` file, I see this:
```
>>> z = zipfile.ZipFile(StringIO(), "w")
>>> z.write("tox.ini")
Exception ignored in: <function ZipFile.__del__ at 0x1076fe7b8>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1789, in __del__
    self.close()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1807, in close
    self._write_end_record()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1911, in _write_end_record
    self.fp.write(endrec)
TypeError: string argument expected, got 'bytes'
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1743, in write
    with open(filename, "rb") as src, self.open(zinfo, 'w') as dest:
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1470, in open
    return self._open_to_write(zinfo, force_zip64=force_zip64)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1583, in _open_to_write
    self.fp.write(zinfo.FileHeader(zip64))
TypeError: string argument expected, got 'bytes'
```

mdickinson:
And this also shows that unfortunately this code isn't being exercised by unit tests. Does it look possible to write a regression test? (And one that would catch the accidental misuse of `StringIO` instead of `BytesIO` for the zip files?)

rahulporuri:
@mdickinson i was going to say that writing tests will take too long but it didn't really. can you take another look at the PR? I _think_ i've addressed all of the comments and i've added the regression tests (to atleast catch trivial bugs).

rahulporuri:
sigh. im not sure what's going on with the CI. all of the linux jobs failed with the following errors, errors which i saw earlier in #92 and didn't investigate - 

```

FAIL: Test if relative paths are set correctly.

  File "/home/travis/build/enthought/apptools/apptools/persistence/tests/test_file_path.py", line 48, in test_relative

    self.assertEqual(f.rel_pth, expect)

AssertionError: 'tmpaEZMUG/data/t.vtk' != '../tmp/tmpaEZMUG/data/t.vtk'

```

apptools/logger/agent/attachments.py
```diff
@@ -58,20 +58,20 @@ def _attach_directory(self, dir):
         relpath = os.path.basename(dir)
 
         import zipfile
-        from cStringIO import StringIO
+        from six import BytesIO
```
mdickinson: Could probably use io.BytesIO here instead of six.BytesIO.

apptools/logger/agent/tests/test_attachments.py
```diff
@@ -0,0 +1,65 @@
...
+from apptools.logger.agent.attachments import Attachments
+
+
+class AttachmentsTestCase(unittest.TestCase):
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, self.tmpdir)
+
+        self.tmpfile = os.path.join(self.tmpdir, "dummy_file.txt")
+        with io.open(self.tmpfile, 'w') as filehandle:
```
mdickinson: Let's use an encoding here, just to follow good practices.

apptools/logger/agent/tests/test_attachments.py
```diff
@@ -0,0 +1,65 @@
...
+        attachments = Attachments(
+            application=MockedApplication(),
+            message=MIMEMultipart()
+        )
+        attachments.package_workspace()
+
+        message = attachments.message
+        self.assertTrue(message.is_multipart())
+        payload = message.get_payload()
+        self.assertTrue(len(payload), 1)
```
mdickinson: `self.assertEqual`?

apptools/logger/agent/tests/test_attachments.py
```diff
@@ -0,0 +1,65 @@
...
+        attachments = Attachments(
+            application=MockedApplication(),
+            message=MIMEMultipart()
+        )
+        attachments.package_single_project()
+
+        message = attachments.message
+        self.assertTrue(message.is_multipart())
+        payload = message.get_payload()
+        self.assertTrue(len(payload), 1)
```
mdickinson: self.assertEqual

apptools/logger/plugin/logger_service.py
```diff
@@ -1,9 +1,11 @@
 # Standard library imports
-from cStringIO import StringIO
 import logging
 import os
 import zipfile
 
+# Third-party library imports
+from six import BytesIO
```
mdickinson: Suggest using `io.BytesIO` instead of `six.BytesIO` here; then we don't need to change this when we drop Python 2 support.

apptools/logger/plugin/tests/test_logger_service.py
```diff
@@ -0,0 +1,43 @@
...
+class LoggerServiceTestCase(unittest.TestCase):
+    def test_create_email_message(self):
+        logger_service = LoggerService()
+        with mock.patch.object(logger_service, 'whole_log_text')\
+                as mocked_log_txt:
+            mocked_log_txt.return_value = "Dummy log data"
+            msg = logger_service.create_email_message(
+                fromaddr='', toaddrs='', ccaddrs='', subject='', priority=''
+            )
+        self.assertIsNotNone(msg)
```
mdickinson: I'd drop this, unless there's some a priori reason to expect the msg to be `None`. This check is redundant with the `assertIsInstance` check below.

apptools/logger/plugin/tests/test_logger_service.py
```diff
@@ -0,0 +1,43 @@
...
+
+        logger_service = MockedLoggerService()
+        with mock.patch.object(logger_service, 'whole_log_text')\
+                as mocked_log_txt:
+            mocked_log_txt.return_value = "Dummy log data"
+            msg = logger_service.create_email_message(
+                fromaddr='', toaddrs='', ccaddrs='', subject='', priority='',
+                include_userdata=True,
+            )
+        self.assertIsNotNone(msg)
```
mdickinson: Same comment as above.

apptools/logger/plugin/tests/test_logger_service.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertIsInstance(msg, MIMEMultipart)
```
mdickinson: Missing trailing newline!

PR #98, opened 2019-09-29
CLN: Remove the unnecessary line continuation ...

prabhuramachandran:
... this causes problems when apptools is installed on some Windows
machines leading to pain when installing Mayavi.

### DIFF
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -602,9 +602,8 @@ class StateUnpickler:
         return State(__metadata__=value)
 
     def _handle_file_path(self, value):
-        if (value['class_name'] == 'FilePath') and \
-           ('file_path' in value['module']) and \
-           self.file_name:
+        if ((value['class_name'] == 'FilePath') and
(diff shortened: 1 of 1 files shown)


mdickinson:
Changes LGTM, too, but as @rahulporuri says it would be good to understand why this is an issue. The original code _should_ be perfectly valid, so if it's being rejected then it's likely something else is wrong.

prabhuramachandran:
Sorry I don't have more information but here is a recent ticket on Mayavi: https://github.com/enthought/mayavi/issues/817 that demonstrates the issue.  I've seen folks running into it but am not sure how this ends up happening when apptools is installed on Windows with Python 3.7 on anaconda.

mdickinson:
@prabhuramachandran I'm going to close this PR. Given that we're not longer using 2to3 on master, this change isn't really useful. At this point, what we really need is a release of apptools.

mdickinson:
> At this point, what we really need is a release of apptools.

Followup: apptools 4.5.0 was released Oct 10th 2019.

PR #101, opened 2019-10-09
Add six, numpy and pandas to requirements

### LINKED ISSUE #99: Make sure that optional dependencies are also optional for tests
Currently the tests fail unless a number of 3rd party packages are installed, including NumPy, Pandas and PyTables. Those tests should be skipped instead if the corresponding packages aren't installed.

rahulporuri:
`six` is now a runtime requirement because we removed the use of `2to3` in the package.
`numpy` and `pandas` have been runtime requirements of a few submodules of `apptools` which were never reported.

fixes #99

### DIFF
--- a/apptools/__init__.py
+++ b/apptools/__init__.py
@@ -7,6 +7,9 @@ except ImportError:
     __version__ = 'not-built'
 
 __requires__ = [
-    'traitsui',
     'configobj',
+    'numpy',
+    'pandas',
+    'six',
+    'traitsui',
 ]


corranwebster:
**Strong** recommendation that Pandas is not a runtime requirement but an optional requirement instead (not sure what exactly requires it, but we _don't_ want people having to install by default).

rahulporuri:
> Strong recommendation that Pandas is not a  runtime requirement but an optional requirement instead (not sure what  exactly requires it, but we don't want people having to install by default).

yep yep. exactly @mdickinson 's concerns as well.
updating.

PR #104, opened 2019-10-09
Temporarily change cwd when running tests

mdickinson:
The file path tests fail if the tests are run with a cwd under `/tmp`. This PR provides a workaround: forcing the cwd to be a local repository directory while running those tests.

This issue was causing CI failures in #92.


rahulporuri:
LGTM. Merging.

PR #54, opened 2015-08-20
Python3 support

prabhuramachandran:
This fixes most of the tests on apptools with Python3.  The io.h5 package has a bunch of issues but the others pass.  These changes are needed for Mayavi.  The changes cherry pick commits from Pradyun's PR here: #18.

PR #122, opened 2020-06-11
Remove uses of implements

mdickinson:
PR #54 missed some uses of `implements`, most of which are now `NameError` s. This PR fixes those.


mdickinson:
Self-merging after self-review.

Issue #120, opened 2020-06-01
Drop Python 2 support


mdickinson:
We no longer need to support Python 2; related code can be dropped.

aaronayres35:
What is the status of this issue?  I'm not sure what else needs to be done to say we have officially dropped python 2 support

mdickinson:
It's looking good; getting rid of `six` was the single biggest thing (done in #230).

Taking a quick look, I spotted a couple of leftovers (but maybe there's still a `sweet_pickle` PR pending?)

- there's still one `from __future__ import print_function` in `apptools/sweet_pickle/tests`
- there's an `if sys.version_info[0] >= 3:` in the `apptools/sweet_pickle/versioned_unpickler.py`

mdickinson:
> but maybe there's still a `sweet_pickle` PR pending?

Ah yes: #199. I'll close this, then.

PR #123, opened 2020-06-11
Replace uses of the Unicode trait with Str

mdickinson:
This PR replaces all uses of the deprecated `Unicode` trait type with `Str`.


mdickinson:
I was jumping the gun a bit here. We can't do this until after we've dropped Python 2 support.

rahulporuri:
This PR is ready to go in now because the upcoming release will not support Python 2.

kitchoi:
> This PR is ready to go in now because the upcoming release will not support Python 2.

I am not sure what sequence of events we are going after here for https://github.com/enthought/apptools/issues/120.
e.g. I'd think both `setup.py` and `etstool.py` need to be updated to prevent Python 2 from being used before changes like this PR goes ahead.

rahulporuri:
I think we can merge this after #190 is merged (after fixing merge conflicts of course).

kitchoi (approved):
LGTM

kitchoi:
I will be resolving the conflicts...

apptools/preferences/tests/test_preferences_helper.py
```diff
@@ -192,7 +192,7 @@ class AcmeUIPreferencesHelper(PreferencesHelper):
             width       = Int(50)
             ratio       = Float(1.0)
             visible     = Bool(True)
-            description = Unicode(u'description')
+            description = Str(u'description')
```
mdickinson, line 195: We could go further and drop the `u` prefixes, but that's probably best done in a separate PR.

mdickinson (commented):
@kitchoi Thank you for the changes! I can't officially approve because I opened the PR in the first place, but LGTM.

aaronayres35 (approved):
LGTM

kitchoi:
Thank you all!

PR #125, opened 2020-06-12
Fix uses of print statement

mdickinson:
This PR fixes some remaining uses of the Python 2 `print` statement.

For now, the source should remain compatible with both Python 2 and Python 3, but some of these `print`s will have changed behaviour in Python 2, printing a representation of a tuple instead of space-separated elements. The Python 2 support is going to be short-lived enough that this isn't worth fixing.

### DIFF
--- a/apptools/logger/ring_buffer.py
+++ b/apptools/logger/ring_buffer.py
@@ -44,10 +44,10 @@ class RingBufferFull:
 # sample of use
 """x=RingBuffer(5)
 x.append(1); x.append(2); x.append(3); x.append(4)
-print x.__class__,x.get()
+print(x.__class__,x.get())
 x.append(5)
-print x.__class__,x.get()
+print(x.__class__,x.get())
 x.append(6)
-print x.data,x.get()
+print(x.data,x.get())
 x.append(7); x.append(8); x.append(9); x.append(10)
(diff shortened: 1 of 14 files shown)


apptools/logger/ring_buffer.py
```diff
@@ -44,10 +44,10 @@ def get(self):
 # sample of use
 """x=RingBuffer(5)
 x.append(1); x.append(2); x.append(3); x.append(4)
-print x.__class__,x.get()
+print(x.__class__,x.get())
```
mdickinson, line 47: This is an example of something that's going to behave differently under Python 2, printing a tuple. But given that Python 2 is mostly dead and that this is commented-out code, it's hard to care. (If we do care, the right fix is to do `from __future__ import print_function` at the top of the module.)

examples/appscripting/example_script_window.py
```diff
@@ -13,6 +13,8 @@
 #------------------------------------------------------------------------------
 
 
+from __future__ import print_function
```
mdickinson, line 16: We need this for the `end = ""` to work, else we'll have a syntax error under Python 2.

mdickinson:
Self-merging after self-review.

PR #131, opened 2020-08-24
Pin pytables to prevent osx segfaults

rahulporuri:
Until we figure out a better solution or we get a new build of pytables, we need to pin pytables to prevent ci/osx from crashing.  https://github.com/enthought/traitsui/issues/1156

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -94,7 +94,9 @@ dependencies = {
     "traitsui",
     "configobj",
     "coverage",
-    "pytables",
+    # This version is pinned due to the latest build causing failures on the
+    # travis CI see issue traitsui/#1156.
+    "pytables==3.5.1-4",
     "pandas",
     "pyface",
     "nose",


rahulporuri:
essentially the same changes as in https://github.com/enthought/traitsui/pull/1155/

etstool.py
```diff
@@ -94,7 +94,9 @@
     "traitsui",
     "configobj",
     "coverage",
-    "pytables",
+    # This version is pinned due to the latest build causing failures on the
+    # travis CI see issue traitsui/#1156.
```
kitchoi, line 98: The comment makes it look like the issue was caused by traitsui, but it isn't. Could we remove the reference please? I think it is okay to reference buildsystem if you must reference something.

kitchoi (approved):
LGTM. One minor comment on the comment :meta:

rahulporuri:
closing as we're going with the alternative approach of moving OSX CI from Travis to Appveyor.

mdickinson:
> closing as we're going with the alternative approach of moving OSX CI from Travis to Appveyor.

Do we know whether the Travis problems are specific to travis-ci.org (as opposed to travis-ci.com)? If so, it may not be worth the effort to move the macOS builds to Appveyor.

kitchoi:
> Do we know whether the Travis problems are specific to travis-ci.org (as opposed to travis-ci.com)?

I don't know.

On the other hand, apptools CI does not seem to be affected by the said problem with pytables. Recent PRs have their CI build passing without an issue... or maybe I am overlooking something again. Persumably we can move only when we see the said segfault, I haven't seen it yet.

rahulporuri:
> Do we know whether the Travis problems are specific to travis-ci.org (as opposed to travis-ci.com)?

We observed the same problem on travis-ci.com so we still needed to move the macos jobs to appveyor.

rahulporuri:
> On the other hand, apptools CI does not seem to be affected by the said problem with pytables.

It did effect the CI jobs and macos jobs got moved in the PR - https://github.com/enthought/apptools/pull/148

kitchoi:
>> On the other hand, apptools CI does not seem to be affected by the said problem with pytables.
>
> It did effect the CI jobs and macos jobs got moved in the PR - #148

Ah I see. Thanks. I did not realize it has already been moved.

PR #148, opened 2020-10-26
Add source click option used on cron jobs - only use python 3.6 on CI - move OSX CI to appveyor

### LINKED ISSUE #152: Travis CI OSX builds fail because of new PyTables EDM eggs
`pytables` is a dependency of `apptools` and we are beginning to see OSX build failures on Python 3.6. This is the same issue as seen earlier in https://github.com/enthought/traitsui/issues/1156, the solution to which is to move OSX CI from Travis to Appveyor. See solution in PR https://github.com/enthought/traitsui/pull/1160.

Build failure first observed in https://github.com/enthought/apptools/pull/148.

aaronayres35:
This PR makes it so that cron jobs install pyface, traitsui, and traits from source by adding a --source click option used in the install function in etstool.py.
It also removes the python 2.7 and 3.5 builds from CI as we will be dropping their support soon.

Additionally, it fixes #152 my moving the OSX CI onto appveyor and off of travis. 

This cron jobs follow what is done in other ets projects (e.g https://github.com/enthought/envisage/blob/8245931f7e3dfcb6513bfb7dc2ae9b10ca6f89a1/etstool.py#L255-L273)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,11 @@ env:
   global:
     - INSTALL_EDM_VERSION=2.0.0
       PYTHONUNBUFFERED="1"
-    - TRAVIS_EVENT_TYPE='cron' 
 
 matrix:
   include:
     - env: RUNTIME=3.6
-    - os: osx
-      env: RUNTIME=3.6
-    fast_finish: true
+  fast_finish: true
 
 cache:
   directories:
@@ -23,7 +20,6 @@ cache:
 before_install:
   - mkdir -p "${HOME}/.cache/download"
   - if [[ ${TRAVIS_OS_NAME} == 'linux' ]]; then ./install-edm-linux.sh; export PATH="${HOME}/edm/bin:${PATH}"; fi
-  - if [[ ${TRAVIS_OS_NAME} == 'osx' ]]; then ./install-edm-osx.sh; export PATH="${PATH}:/usr/local/bin"; fi
   - edm install -y wheel click coverage
 install:
   - if [[ ${TRAVIS_EVENT_TYPE} == 'cron' ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,10 @@ before_install:
   - if [[ ${TRAVIS_OS_NAME} == 'osx' ]]; then ./install-edm-osx.sh; export PATH="${PATH}:/usr/local/bin"; fi
   - edm install -y wheel click coverage
 install:
-  - edm run -- python etstool.py install --runtime=${RUNTIME} || exit
   - if [[ ${TRAVIS_EVENT_TYPE} == 'cron' ]]; then
-      edm run -- python etstool.py install ${RUNTIME} --source || exit;
+      edm run -- python etstool.py install --runtime=${RUNTIME} --source || exit;
+    else
+      edm run -- python etstool.py install --runtime=${RUNTIME} || exit;
     fi
 script:
   - edm run -- python etstool.py test --runtime=${RUNTIME} || exit
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,7 @@ env:
 
 matrix:
   include:
-    - env: RUNTIME=2.7
-    - env: RUNTIME=3.5
     - env: RUNTIME=3.6
-    - os: osx
-      env: RUNTIME=2.7
-    - os: osx
-      env: RUNTIME=3.5
     - os: osx
       env: RUNTIME=3.6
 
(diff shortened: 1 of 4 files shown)


rahulporuri:
fixes #128

rahulporuri (approved):
LGTM. I tested this locally and it works as expected - but I needed to explicitly specify the runtime as 3.6 because 3.5 as the default runtime.

Like I mentioned in the enable PR, you might be able to set the env var `TRAVIS_EVENT_TYPE` in the env vars section of the travis config file.

aaronayres35:
CI is failing on OSX and python 3.6 with 
```
Command '['edm', 'run', '-e', 'apptools-test-3.6', '--', 'coverage', 'run', '-p', '-m', 'nose.core', '-v', 'apptools', '--nologcapture']' returned non-zero exit status 252.
```

aaronayres35:
> CI is failing on OSX and python 3.6 with
> 
> ```
> Command '['edm', 'run', '-e', 'apptools-test-3.6', '--', 'coverage', 'run', '-p', '-m', 'nose.core', '-v', 'apptools', '--nologcapture']' returned non-zero exit status 252.
> ```

Note this is happening without `${TRAVIS_EVENT_TYPE} == 'cron'` so it seems potentially unrelated to this PR?

aaronayres35:
Ah, we should only be installing from source for python 3.6 correct? All the other builds with different python versions fail but I believe that is to be expected?

e.g. for the python 3.5 build there is the error:
```
traits requires Python '>=3.6' but the running Python is 3.5.2
```

I will set up cron job to only run on 3.6

aaronayres35:
OSX on python 3.6 fails with:
```
Command '['edm', 'run', '-e', 'apptools-test-3.6', '--', 'coverage', 'run', '-p', '-m', 'nose.core', '-v', 'apptools', '--nologcapture']' returned non-zero exit status 252.
```

aaronayres35:
Also this is currently a non-ideal solution.  The current code is:
```
- edm run -- python etstool.py install --runtime=${RUNTIME} || exit
  - if [[ ${TRAVIS_EVENT_TYPE} == 'cron' ]]; then
      edm run -- python etstool.py install --runtime=3.6 --source || exit;
    fi
```
So on a crob job with runtime 2.7 for example, we run `edm run -- python etstool.py install --runtime=${RUNTIME} || exit` which sets up the `apptools-test-2.7`. Then the cron job if statement sets up the `apptools-test-3.6` and then the tests get run in the `apptools-test-2.7` environment.  This just means that on crob jobs the 2.7 and 3.5 runtime builds are run normally which is fine, but the `apptools-test-3.6` environment setup/install is unnecessary.

rahulporuri:
two things : 

- given that we will be dropping python 2.7 and 3.5 support soon, it's safe to remove them from the CI matrix immediately, and not worry about scenarios involving them
- i think the osx issue is related to pytables (which is a dependency of apptools) which we saw in traitsui (issue - https://github.com/enthought/traitsui/issues/1156, PR - https://github.com/enthought/traitsui/pull/1160). You should be able to confirm this by enabling faulthandler when running the testsuite. If I am correct, you will need to update OSX CI to use appveyor instead of travis.

rahulporuri:
i opened the issue https://github.com/enthought/apptools/issues/152 regarding moving OSX CI from Travis to Appveyor.

.travis.yml
```diff
@@ -7,18 +7,14 @@ env:
   global:
     - INSTALL_EDM_VERSION=2.0.0
       PYTHONUNBUFFERED="1"
+    - TRAVIS_EVENT_TYPE='cron'
```
rahulporuri, line 10:
```suggestion

```

.travis.yml
```diff
@@ -10,15 +10,10 @@ env:
...
-    - env: RUNTIME=2.7
-    - env: RUNTIME=3.5
     - env: RUNTIME=3.6
-    - os: osx
-      env: RUNTIME=2.7
-    - os: osx
-      env: RUNTIME=3.5
     - os: osx
       env: RUNTIME=3.6
+    fast_finish: true
```
rahulporuri, line 16:
```suggestion
    - os: osx
      env: RUNTIME=3.6
  fast_finish: true
```

looks like `fast_finish` needs to be at the `matrix` level, not `include` level.

aaronayres35:
> * i think the osx issue is related to pytables (which is a dependency of apptools) which we saw in traitsui (issue - [enthought/traitsui#1156](https://github.com/enthought/traitsui/issues/1156), PR - [enthought/traitsui#1160](https://github.com/enthought/traitsui/pull/1160)). You should be able to confirm this by enabling faulthandler when running the testsuite. If I am correct, you will need to update OSX CI to use appveyor instead of travis.

faulthandler is giving same traceback that was seen in CI on this commit https://github.com/enthought/traitsui/pull/1155/commits/e0b285f9c6e0ddd10a09cc0d08c2bfa3eb9d5b9d 
for the pytables issue seen on traits ui.  

Going to go ahead and update OSX CI to use appveyor instead of travis as it was done in the PR you mentioned

rahulporuri (approved):
still LGTM

PR #155, opened 2020-10-29
Use Python 3.6 as the default runtime in etstool module

aaronayres35:
fixes #149 
by simply bumping 3.5 to 3.6 in click options.

PR #156, opened 2020-10-30
Default runtime should be 3.6, not 6

rahulporuri:
Bug introduced in PR https://github.com/enthought/apptools/pull/155/. I should have caught the bug because I reviewed the PR 😅

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -84,6 +84,8 @@ from contextlib import contextmanager
 
 import click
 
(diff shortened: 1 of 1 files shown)


kitchoi (approved):

kitchoi:
There probably should be a module level constant for this, e.g.
https://github.com/enthought/traitsui/blob/c5438d56a0c9f53e03405cb0450d5b0f7c78875e/etstool.py#L102-L103

Issue #119, opened 2020-02-12
Convert test suite to use unittest


mdickinson:
The test suite is currently based on nose. nose is no longer maintained. There's a successor, `nose2`, but that doesn't pick up all the apptools tests right now. (Out of the box, nose finds 401 tests, while nose2 finds 174. That may be tweakable.)

We should consider converting all the tests to regular unittest testcases (and then users can still choose to use whichever test runner they prefer).

aaronayres35:
As mentioned in #153 (closed as a duplicate), we will also want to use unittest as the test runner in etstool.py

https://github.com/enthought/apptools/blob/5a136f642d6865970497aa3b2f1f176d56841de2/etstool.py#L146-L147

PR #158, opened 2020-10-30
Rename test modules

rahulporuri:
This PR renames the test modules from `*_test_case` to `test_*`. This ensures that the test modules can be discovered using unittest instead of nose. Additionally, this PR removes the `if __name__ == "__main__"` code block from unittest test modules.


rahulporuri:
this is one small step towards #119

kitchoi (approved):

PR #160, opened 2020-10-30
Use unittest instead of nose for test discovery

### LINKED ISSUE #119: Convert test suite to use unittest

rahulporuri:
Along with the PRs #159 and #158 , this PR fixes #119 .

This PR updates the testsuite such that the entire test suite is discoverable by the unittest test discovery mechanism instead of the nose test discovery mechanism.

The main change is to refactor test functions into test methods on a unittest `TestCase` classes. Additionally, a few missing `__init__.py` modules were added to ensure that the unittest discovery mechanism actually discovers the tests.

Finally, this PR removes `nose` from the dev env and replaces nose with unittest in the `test` click command in the `etstool` helper module.

### DIFF
--- a/apptools/io/h5/tests/test_dict_node.py
+++ b/apptools/io/h5/tests/test_dict_node.py
@@ -1,3 +1,5 @@
+import unittest
+
 import numpy as np
 from numpy.testing import raises, assert_allclose
 import six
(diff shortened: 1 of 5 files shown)


mdickinson:
I see that `nose` was added back in. Do we still need `nose` if we replace uses of `raises` with uses of `assertRaises`?

mdickinson:
Closing and re-opening to trigger migrated CI (travis-ci.org CI was migrated to travis-ci.com CI)

mdickinson:
Closing and re-opening one more time: I was expecting the last CI run to use travis-ci.com, but it's still using travis-ci.org.

rahulporuri:
> I see that `nose` was added back in. Do we still need `nose` if we replace uses of `raises` with uses of `assertRaises`?

No we won't. I'll attempt to replace `raises`.

mdickinson (approved):
LGTM

PR #165, opened 2020-11-02
Fix assert deprecation and resource warnings in testsuite

### LINKED ISSUE #162: Assert deprecation and resource warnings reported when running the testsuite
- The `TestCase.assert_` assertion has been deprecated in favor of `TestCase.assertTrue`
- `ResourceWarning` reported because of unclosed file.

aaronayres35:
fixes #162 

This PR simply replaces all occurrences of `self.assert_(` with `self.assertTrue(` to avoid deprecation warnings, and also avoids a ResourceWarning for not closing a file by using a `with` statement when opening the file.  

One thing I noticed when making the `assertTrue` fix was that there are tons of places which use 
`assertEqual(____, True)` or `assertEqual(____, None)` or `assertEqual(____, False)`.  Rather than just `assertTrue`, `assertIsNone` or `assertFalse`.  Similarly there are many places were we have `assertTrue(isinstance(____,____))` instead of `assertIsInstance(__, ___)` etc.  These are extremely minor, just something we may want to clean up

### DIFF
--- a/apptools/io/tests/test_file.py
+++ b/apptools/io/tests/test_file.py
@@ -161,7 +161,8 @@ class FileTestCase(unittest.TestCase):
         # Create the file.
         f.create_file(content)
         self.assertEqual(f.exists, True)
-        self.assertEqual(open(f.path).read(), content)
+        with open(f.path) as file:
+            self.assertEqual(file.read(), content)
 
         # Try to create it again.
         self.assertRaises(ValueError, f.create_file, content)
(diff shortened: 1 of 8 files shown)


rahulporuri (approved):
LGTM. I checked the CI logs and I dont see any warnings.

rahulporuri:
> One thing I noticed when making the assertTrue fix was that there are tons of places which use
assertEqual(____, True) or assertEqual(____, None) or assertEqual(____, False). Rather than just assertTrue, assertIsNone or assertFalse. Similarly there are many places were we have assertTrue(isinstance(____,____)) instead of assertIsInstance(__, ___) etc. These are extremely minor, just something we may want to clean up

Can you address them in a separate PR? I'm guessing that it wont take more than an hour to complete this. If not, can you open an issue?

PR #166, opened 2020-11-02
update example.ini so it is not modified by the test suite

### LINKED ISSUE #118: Test suite modifies a tracked file
When I run the test suite under Python 3.8.1 (on macOS, in a venv), it modifies the file `apptools/preferences/tests/example.ini`:

```
(apptools-3.8) mirzakhani:apptools mdickinson$ git diff
diff --git a/apptools/preferences/tests/example.ini b/apptools/preferences/tests/example.ini
index 100dc6e1..c535635c 100644
--- a/apptools/preferences/tests/example.ini
+++ b/apptools/preferences/tests/example.ini
@@ -3,7 +3,7 @@ bgcolor = blue
 width = 50
 ratio = 1.0
 visible = True
-description = 'acme ui'
+description = acme ui
 offsets = "[1, 2, 3, 4]"
 names = "['joe', 'fred', 'jane']"
 
```

Steps to reproduce, from a clean clone:

- Create a Python 3.8.1 venv, and activate it (`python3.8 -m venv ~/.venvs/apptools-3.8`, `source ~/.venvs/apptools-3.8/bin/activate`
- Install dependencies: `pip install nose numpy tables pandas`
- Install apptools: `pip install -e .`
- `git status` at this point shows that the working tree is clean
- Run tests: `nosetests .`
- `git status` now shows modifications to `examples.ini`.

aaronayres35:
fixes #118 

I found that the cause of the modification was here:
https://github.com/enthought/apptools/blob/faabb374405c366e7e3bd33745824d6e64ff1f69/apptools/preferences/tests/test_preferences_helper.py#L211-L257

In the test, the description is modified, reset and saved.  However, when it is reset it is set to the string it was originally, but when saving that string simply gets written to the `example.ini` file as the content of the string, without the quotation marks.  This seems to be the expected behavior based on looking at and messing with this other test:
https://github.com/enthought/apptools/blob/faabb374405c366e7e3bd33745824d6e64ff1f69/apptools/preferences/tests/test_preferences.py#L429-L475


kitchoi:
Hmmm I think while the current change fixes the observed behaviour that the file content is changed after running the test, I suspect the right fix here is to avoid writing to the example file from the test, because the file is a fixture used by other tests. e.g. The test can save the content to a temporary file. It is common to create a temporary directory for this kind of test.

(Orthogonal: looks like the `save` method being used also accepts a file-like object, so we could also do all these manipulations in memory without using the file system, e.g. using `StringIO `...)

apptools/preferences/tests/test_preferences_helper.py
```diff
@@ -53,11 +55,17 @@ def setUp(self):
...
+        # A temporary directory that can safely be written to.
+        self.tmpdir = tempfile.mkdtemp()
+
         return
 
     def tearDown(self):
         """ Called immediately after each test method has been called. """
 
+        # Remove the temporary directory.
+        os.removedirs(self.tmpdir)
```
mdickinson, line 67: See #178; this should be changed to `rmdir` (though the solution I generally prefer is to use `shutil.rmtree`; that way you don't have to remove anything that was created inside the temporary directory first).

apptools/preferences/tests/test_preferences_helper.py
```diff
@@ -245,16 +253,22 @@ class AcmeUIPreferencesHelper(PreferencesHelper):
...
+        try:
+            # Load it into a new node.
+            p = Preferences()
+            p.load(tmp)
+            self.assertEqual(second_unicode_str, p.get('acme.ui.description'))
+            self.assertEqual(u'True', p.get('acme.ui.visible'))
+            self.assertTrue(helper.visible)
+        finally:
+            # Clean up!
+            os.remove(tmp)
```
kitchoi, line 269: This is why we more often see `shutil.rmtree` in the tear down, so that you don't have to make the test code more complicated by making sure each file is removed. e.g. We won't need this `os.remove` here if teardown is calling `shutil.rmtree`. All files in the temporary directory will be deleted no matter what. If you don't mind making this one more change to use `rmtree`, it would make writing new tests easier in the future.

kitchoi (approved):
LGTM

apptools/preferences/tests/test_preferences_helper.py
aaronayres35: yep, sure thing

kitchoi:
Thank you @aaronayres35 
Confirmed that the example file is no longer modified after running the tests (the modified timestamp reported by the OS has not changed).

mdickinson:
@aaronayres35 Many thanks for tracking this down and fixing it. It had been bugging me every single time I ran the `apptools` test suite, but never quite enough to go and track down the cause.

PR #167, opened 2020-11-02
TST: Switch on default warning flag for CI test command

### LINKED ISSUE #150: Switch on the default warning flag when running testsuite
See related https://github.com/enthought/envisage/issues/311

Deprecation warnings triggered during test runs are visible, but deprecation warnings triggered prior to when a unittest test case is loaded are by default hidden from the console.

With that, deprecation warnings from trait types definition do not show up from running the current test command, because the warnings occur at import time.

aaronayres35:
Fixes #150 

This change is lifted directly from https://github.com/enthought/traitsui/pull/1059.  It simply adds `-W default` flag.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -181,8 +181,8 @@ def test(runtime, environment):
     environ = {}
     environ['PYTHONUNBUFFERED'] = "1"
     commands = [
-        ("edm run -e {environment} -- python -W default -m coverage run -p -m "
-         "unittest discover -v apptools")
+        "edm run -e {environment} -- python -W default -m coverage run -p -m "
+         "unittest discover -v apptools"
     ]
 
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -181,8 +181,8 @@ def test(runtime, environment):
     environ = {}
     environ['PYTHONUNBUFFERED'] = "1"
     commands = [
-        ("edm run -e {environment} -- python -W default -m coverage run -p -m "
-         "unittest discover -v apptools")
+        "edm run -e {environment} -- python -W default -m coverage run -p -m "
+         "unittest discover -v apptools"
```
rahulporuri, line 185: nit-pick : indentation is off by one.

rahulporuri (approved):
LGTM except for one small nit-pick.

PR #169, opened 2020-11-03
Updating assertion intent in tests

aaronayres35:
This PR updates varying assert statements to be more specific.  e.g. it replaces things like `self.assertEqual(True, isinstance(a,b))` with `self.assertIsInstance(a,b)`.

see comment: https://github.com/enthought/apptools/pull/165#issuecomment-720962048 for the original motivation.


mdickinson:
Just a drive-by comment: note that `self.assertEqual(some_object, True)` is a strictly stronger test than `self.assertTrue(some_object)` - for the former,  `some_object` must be either `True` or a number with value `1`, while for the latter, the object simply has to be truthy.

For the vast majority of these changes, this likely doesn't matter, but it's possible that there are cases where we want to keep the stronger test.

kitchoi:
> Just a drive-by comment: note that self.assertEqual(some_object, True) is a strictly stronger test than self.assertTrue(some_object) - for the former, some_object must be either True or a number with value 1, while for the latter, the object simply has to be truthy.

Agreed. That being said, I think a lot of `assertTrue` and `assertFalse` are applied on boolean values where we expect user code to be doing `if x:` or `while x:` (etc). If that is how the values are intended to be used, it would make sense to test the truthy behaviour.

kitchoi:
Perhaps the title and the description of the PR should be updated to say something along the line of "updating assertion intent in tests"?  As Mark said, the switch from assertEqual to assertTrue is actually less "specific" / "strict".

apptools/preferences/tests/example.ini
```diff
@@ -3,7 +3,7 @@ bgcolor = blue
 width = 50
 ratio = 1.0
 visible = True
-description = 'acme ui'
```
kitchoi, line 6: This looks like an accidental change from #166, which is orthogonal.

apptools/preferences/tests/test_py_config_file.py
```diff
@@ -49,7 +49,7 @@ def test_load_from_filename(self):
         self.assertEqual('blue', config['acme.ui']['bgcolor'])
         self.assertEqual(50, config['acme.ui']['width'])
         self.assertEqual(1.0, config['acme.ui']['ratio'])
-        self.assertEqual(True, config['acme.ui']['visible'])
+        self.assertTrue(config['acme.ui']['visible'])
```
kitchoi, line 52: This might be worth keeping the original `assertEqual(True, ...)`.  This is loading data from the config file, and it is worth making sure this is actually a bool, not, e.g. `"true"`.

apptools/io/tests/test_folder.py
```diff
@@ -108,77 +108,77 @@ def test_properties(self):
...
+        self.assertFalse(g.exists)
 
         return
 
     def test_create_folder(self):
         """ folder creation """
 
         f = File('data/sub')
-        self.assertEqual(f.exists, False)
+        self.assertFalse(f.exists, False)
```
mdickinson, line 177: The second argument should be removed here (or replaced with a suitable message, if you're feeling adventurous).
```suggestion
        self.assertFalse(f.exists)
```

apptools/preferences/tests/test_py_config_file.py
```diff
@@ -68,7 +68,7 @@ def test_load_from_file(self):
...
```
kitchoi, line 71: Same here, and the ones below to do with config.

apptools/io/tests/test_folder.py
```diff
@@ -108,77 +108,77 @@ def test_properties(self):
...
 
         # Create the folder.
         f.create_folder()
-        self.assertEqual(f.exists, True)
+        self.assertTrue(f.exists, True)
```
mdickinson, line 181:
```suggestion
        self.assertTrue(f.exists)
```

apptools/preferences/tests/test_preferences_helper.py
```diff
@@ -92,7 +92,7 @@ class AcmeUIPreferencesHelper(PreferencesHelper):
         self.assertEqual('blue', helper.bgcolor)
         self.assertEqual(50, helper.width)
         self.assertEqual(1.0, helper.ratio)
-        self.assertEqual(True, helper.visible)
+        self.assertTrue(helper.visible)
```
kitchoi, line 95: I could not find an assertion in these tests that check `assertFalse(helper.visible)`.  `helper.visible` could be a method instead of a scalar value, and that would pass this `assertTrue` check.
```
>>> class A:
...     def visible(self):
...         return False
... 
>>> bool(A().visible)
True
```

mdickinson:
> it's possible that there are cases where we want to keep the stronger test.

I just looked through all these; they all look good to me - I can't find any cases where we actually want to assert equality with `True` or `False`.

mdickinson (approved):
LGTM modulo a couple of small changes - I've left suggestions for those.

mdickinson:
> it would make sense to test the truthy behaviour.

Yes, absolutely. It's the right change to make in almost all cases; I'm just saying that we need to be on the alert for corner cases where we really do want to check equality with `True`. But I didn't find any of those. :-)

kitchoi (commented):
Some of the changes make the assertion more Pythonic, and may well have been the original intended behaviour: e.g. `File.exists` is likely meant to be used like this:
```
some_file = File(...)
if some_file.exists:
    ...
```
Then if I was to write the test from scratch, I would have used `assertTrue(some_file.exists)` too.

Some changes have, technically speaking, made the assertions stricter, like `assertEqual(None, ...)` to `assertIsNone(...)`. There exist libraries that override `__eq__` so that it can be compared with `None` (Don't do that, please!). But that is likely not applicable here...

The distinction between `assertEqual`, `assertTrue`, `assertIs(..., True)` is subtle, and sometimes we do care about those fine differences. Consider this:
```
>>> 1 == True
True
```
The value might need to be serialized to a file and be loaded back, like the config file. In those situations, we care about the exact type of the value.

All in all, I bet most of the changes here are probably okay, but I do have to apply some elements of guessing. While some of the original assertions may have been an over-specification and the changes make the intention more clear, the original assertions are not causing much trouble either. Maybe we shall consider closing this as not-worth-it, or just keep the changes to `assertIsNone`, `assertIsInstance` (the identity checks), which are obviously correct.

kitchoi:
Ahahaha @mdickinson beat me to it, oops :D

apptools/preferences/tests/test_preferences_helper.py
aaronayres35: Should I revert this change?  In all the `AcmeUIPreferencesHelper` class definitions, `visible` is specified as a `Bool`

kitchoi: Sorry. Never mind this one then.

kitchoi (approved):
LGTM too!

PR #170, opened 2020-11-03
Require pandas on the only method that requires it

kitchoi:
Currently `apptools.io.h5` requires `pandas` to be a runtime requirement because `DataFrame` is imported and then used ONCE in `H5TableNode.to_dataframe`.

The `to_dataframe` merely instantiates `DataFrame` with the public attribute `ix` and return that DataFrame. It is fair to assume that if one requires this `to_dataframe`, they should already have pandas in their environment. It is a bit overkill to require pandas to be installed, especially if `to_dataframe` is not actually required in user code.

This PR moves the import of `pandas` in the single method that requires it. And fixes the tests for H5TableNode so that they are skipped if NumPy and PyTables are not available.  The latter contributes towards #99 (but it does not close the issue).

### DIFF
--- a/apptools/_testing/__init__.py
+++ b/apptools/_testing/__init__.py
@@ -1,0 +1,12 @@
+# (C) Copyright 2006-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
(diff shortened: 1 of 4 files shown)


rahulporuri (approved):
LGTM

corranwebster:
Do you want to add an `extras_require` to the `setup.py` to allow pip-install of pandas as an optional dependency?

kitchoi:
> Do you want to add an extras_require to the setup.py to allow pip-install of pandas as an optional dependency?

Good idea. Opened https://github.com/enthought/apptools/issues/176

PR #172, opened 2020-11-03
Remove appscripting subpackage, its docs and examples, and references to it

### LINKED ISSUE #132: Remove appscripting subpackage
There are two application scripting systems in AppTools.  Mayavi uses the `apptools.scripting` package.  No Enthought application has used `apptools.appscripting` in the last 10 years or so.  Therefore `apptools.scripting` should be removed.

This removal should probably occur in the next major release, with any third party users encouraged to use `apptools.scripting`.

aaronayres35:
fixes #132 

This PR deletes the entire `appscripting` subpackage along with its related docs and examples.  Additionally, it removes any other references to it inside `apptools`


rahulporuri (approved):
LGTM. I was wondering if any part of the appscripting docs can be repurposed as docs for scripting but I doubt it.

aaronayres35:
> LGTM. I was wondering if any part of the appscripting docs can be repurposed as docs for scripting but I doubt it.

In any case if any of them can be repurposed, we can resurrect them in a separate PR

PR #173, opened 2020-11-03
Remove template subpackage and any references to it

### LINKED ISSUE #140: Remove apptools.template.test.enable_editor
This has been superseded by Enable's component editor.

### LINKED ISSUE #141: Decide fate of apptools.template
The `apptools.template` package is designed to provide a system for configuring traits classes and instantiating them according to the template.  This is motivated by, for example, allowing a user to create templates to style Chaco plots and then use those style choices when creating the actual plot instances.

The template objects are also designed to help with serialization and deserialization, allowing the configuration of application state to be separated from the data sources when serializaing.

This is potentially a very useful set of functionality, but it may not be worth preserving.

Or, alternatively, apptools is not the place for it, and it really needs to live in Chaco until a wider use for it is found.

aaronayres35:
fixes #141 
fixes #140 

The fate of the `template` subpackage has been decided, and it was decided that it should be removed entirely. 
This PR deletes the `template` subpackage and removes any references to it from the code base.


TODO.txt
```diff
@@ -1,7 +1,3 @@
-* Take a critical look at whether apptools.template should be in this project.
-  It appears to be adding a lot of dependencies the other packages do not need.
-  For now, its use is controlled through an extra.
-
```
aaronayres35, line 4: This file has not been edited since 2011, and it seems likely that all of these are now obsolete.  Even though it is orthogonal, should I just delete this file with this PR (I can open github issues for any of the bullets if needed, but they appear no longer relevant)?

rahulporuri: let's do that in a separate PR.

rahulporuri (approved):
LGTM. I dont see any documentation for the template submodule so this PR looks complete.

PR #175, opened 2020-11-03
Remove permissions subpackage, its docs and examples, and references to it

### LINKED ISSUE #174: Remove permissions subpackage
It has been decided that the `apptools.permissions` subpackage should be deleted. 

It appears to be unused, has some overlap with an internal closed-source library, and is untested.

aaronayres35:
fixes #174 

This PR deletes the entire `apptools.permissions` subpackage along with its related docs and examples. Additionally, it removes any other references to it inside `apptools`.


rahulporuri (approved):
LGTM.

I'm sure we won't forget when the time comes but just to be safe, can you update the changelog file with entries for this PR and the other two which remove entire submodules?

PR #179, opened 2020-11-04
Don't combine coverage in CI

rahulporuri:
We dont report coverage anyway so there is no point in doing `coverage combine` in CI, it just takes up time unnecessarily. Note that #164 removed coverage reporting in CI.


aaronayres35 (approved):
LGTM!

PR #181, opened 2020-11-04
Add routine to add changelog news fragment

kitchoi:
This PR proposes a small developer workflow to add news fragments for the generation of Changelog.

The idea is that a PR will contribute a small fragment to the changelog, if the changes from the PR is news worthy for the end-users. This will ease the process of having to parse and digest many PRs at the time of making a release. That manual process is particular tedious for PRs that are merged around the time of a previous release but not actually included in that release (this is more of an issue for projects that run a maintenance branch and have many PR activities).

Similar routines can be found in CPython, NumPy, SymPy and various other projects. This is a cheap first-cut solution, without nice tooling like GitHub bots etc.

This PR
- Adds GitHub Pull Request template to remind us of news fragments
- Adds two commands in `etstool.py` for creating news fragments and building the changelog from the fragments.
- Adds README for the folder that will contains the fragment (also referred to by the GitHub PR template).

Currently this is an experiment, and maybe be removed or replaced in the future subject to evaluation.

### DIFF
--- a/docs/releases/README.rst
+++ b/docs/releases/README.rst
@@ -1,0 +1,27 @@
+The `upcoming` directory contains news fragments that will be added to the
+changelog for the NEXT release.
+
+Changes that are not of interest to the end-user can skip adding news fragment.
+
+Add a news fragment
+-------------------
+Create a new file with a name like ``<pull-request>.<type>.rst``, where
+``<pull-request>`` is a pull request number, and ``<type>`` is one of:
+
+- ``feature``: New feature
+- ``bugfix``: Bug fixes
+- ``deprecation``: Deprecations of public API
+- ``removal``: Removal of public API
+- ``doc``: Documentation changes.
+
+Then write a short sentence in the file that describes the changes for the
+end users, e.g. in ``123.removal.rst``::
+
+    Remove package xyz.
+
+Alternatively, use the following command, run from the project root directory
+and answer the questions::
+
+    python etstool.py changelog create
+
+(This command requires ``click`` in the environment.)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+
+**Checklist**
+- [ ] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)
--- a/docs/releases/upcoming/116.bugfix.rst
+++ b/docs/releases/upcoming/116.bugfix.rst
@@ -1,0 +1,1 @@
+Fix SyntaxWarning in persistence.file_path (#116)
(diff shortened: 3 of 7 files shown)


docs/releases/README.rst
```diff
@@ -0,0 +1,27 @@
+The `upcoming` directory contains news fragments that will be added to the
+changelog for the NEXT release.
+
+Changes that are not of interest to the end-user can skip adding news fragment.
+
+Add a news fragment
+-------------------
+Create a new file with a name like ``<pull-request>.<type>.rst``, where
```
kitchoi, line 8: The obvious flaw here is that if you try to do this before opening a PR, you don't know the PR number. So you would either have to guess it, or do this step right after opening a PR. This is why having a GitHub bot would be nice, but it is the best we could do with the least effort right now.

aaronayres35 (approved):
LGTM, I tested both the `python etstool.py changelog create` and `python etstool.py changelog build` commands locally and they seem to work as expected.

As you mentioned this is a first-cut generation, so the results still need a little tweaking but that is perfectly fine for now.  Some tweaks could be made so that the new updates get added to the changelog after the 
```
Apptools CHANGELOG
==================
```
at the start of the file.  Similarly the version number could be extracted from `setup.py`.  In any case we don't need to address these minor things now (ignore me).

kitchoi:
Thank you @aaronayres35 

@rahulporuri is this okay to be merged?  We can treat this as a trial and then evaluate (probably after the next release) whether we should keep / change this.

docs/releases/README.rst
rahulporuri: I think this is a step towards making life easier for ourselves so for now, it's fine if PR authors have to add the news fragment after opening the PR.

rahulporuri (approved):
This is awesome :D I tested the two click commands and they work as expected.

Just to be sure, the release process now involves removing the files - which needs to happen in the PR which bumps version to continue development on master branch. Is that correct?

kitchoi:
> Just to be sure, the release process now involves removing the files - which needs to happen in the PR which bumps version to continue development on master branch. Is that correct?

Ah good point. Initially I was thinking of doing the changelog in a PR separate from the one that bump the version. But then I think I was thinking of the maintenance branch + release candidate scenario in Traits, TraitsUI and Pyface where the maintenance branch is created _before_ the release: One would update the changelog and clear the news fragments in the maintenance branch. After a release is made, the updated changelog (including the 'deletion' of new fragments) are merged back to the master. That is where we can avoid missing changes that are merged around a release but not included in the release. The news fragments made for changes on master that are not included in a release will not be affected (I think git hashes/indexes files, not folders).
(e.g. traitsui#1383 was not included in TraitsUI 7.1.0 but was merged before the release PR, so it is at risk of being lost in the next changelog update if one just scans the list of merged PR). 

Your idea of doing the changelog update and the removal of files in the same PR that bumps the version makes sense in the scenario where we are doing the simple single PR release process from master without release candidates.

PR #184, opened 2020-11-04
Remove lru_cache subpackage

### LINKED ISSUE #91: Remove LRUCache from apptools
[Update] A decision has been made to remove `LRUCache` from `apptools`. See comment further down the thread - https://github.com/enthought/apptools/issues/91#issuecomment-722233227

Python 3.x provides lru cache via the functools stdlib package, which is available on Python 2.x as the backported `functools32` package.

I don't know what the differences between the `LRUCache` that `apptools` provides and how it is different from the one from `functools`. I am also not sure who is using `LRUCache` from `apptools`.

Note that if we do choose to remove the module/class, the next release will have to be a major release.

See https://github.com/enthought/apptools/blob/master/apptools/lru_cache/lru_cache.py

### LINKED ISSUE #135: Document apptools.lru_cache
The apptools.lru_cache is a comparatively recent addition to the apptools suite.  It provides a traitified version of an LRU cache.  While this is a fairly well-understood concept, it should have at least minimal documentation and examples of use.

### LINKED ISSUE #147: LRU Cache should implement Mapping interface?
It would be straightforward to add methods so that the `LRUCache` class implements the `Mapping` (or perhaps even `MutableMapping`) interface which would make it useful in a wider context.  The key missing functionality is `__iter__`, which seems like a reasonable thing to be able to ask (ie. iterate over everything in the cache to do something); there are some questions around locking that may need to be sorted out (probably want to lock for the duration of the iteration, but that may be awkward, alternative is to copy and iterate on the copy).

aaronayres35:
closes #91 
closes #147 
closes #135 

It was decided in an offline discussion today that the `lru_cache` sub package can be removed. 
This PR simply removes the package

It was also suggested that we may want to do a quick scream test on slack before pushing this through, just to be safe.


rahulporuri (approved):
LGTM. Did you scream on slack?

aaronayres35:
> LGTM. Did you scream on slack?

Not yet I can now

kitchoi:
Now that #181 is merged, could you add a news fragment please?

PR #186, opened 2020-11-05
Use the latest EDM version (>=3.0.1)

### LINKED ISSUE #154: Use the latest EDM version (>=3.0.1)
A new major version of EDM is available and it would be good to update to the latest version - https://assets.enthought.com/downloads/edm/. This should be as simple as updating the environment variables in travis and appveyor config files.
https://github.com/enthought/apptools/blob/5a136f642d6865970497aa3b2f1f176d56841de2/appveyor.yml#L6
https://github.com/enthought/apptools/blob/5a136f642d6865970497aa3b2f1f176d56841de2/.travis.yml#L8

Note that there is a major behavioral change with version 3.0.1 of EDM - the default platform tag used on unix is rh7, not rh6. Previously, users needed to explicitly pass the rh7 platform cmdline flag if they wanted to created an rh7 environment and rh6 was the default if no flag was set.

We weren't setting this flag earlier -

https://github.com/enthought/apptools/blob/5a136f642d6865970497aa3b2f1f176d56841de2/etstool.py#L123

When we update edm to version 3.0.1, we will automatically switch to using the rh7 environments + rh7 packages in the environments instead of rh6. This should not break anything but we need to be aware of this change.

aaronayres35:
Fixes #154 

This PR simply bumps the version number for the edm version from 2.0.0 to 3.0.1 in the appveyor and travis yml files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 
 env:
   global:
-    - INSTALL_EDM_VERSION=2.0.0
+    - INSTALL_EDM_VERSION=3.0.1
       PYTHONUNBUFFERED="1"
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 
 env:
   global:
-    - INSTALL_EDM_VERSION=3.0.1
+    - INSTALL_EDM_VERSION=3.1.1
       PYTHONUNBUFFERED="1"
 
 matrix:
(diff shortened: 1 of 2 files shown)


mdickinson:
Why not EDM 3.1.1?

aaronayres35:
> Why not EDM 3.1.1?

I forgot about its existence / just followed other ets projects 😅

kitchoi (commented):
LGTM

aaronayres35:
> LGTM

@kitchoi looks like this review got left as a comment not an approval.  When you get a chance could you approve it please?

kitchoi (approved):
> looks like this review got left as a comment not an approval. When you get a chance could you approve it please?

Oops, sorry about that!

PR #187, opened 2020-11-05
Dont import unittest from traits

rahulporuri:
This PR updates `unittest` imports to use the `unittest` standard library package directly instead of importing from `traits` testing utility. This might have been done earlier to provide compatibility between unittest2 on python 2 and unittest on python 3 but now that we are python 3 only, this is not needed.

Note additionally that the traits module has long been updated to only export the `unittest` standard library - https://github.com/enthought/traits/blob/master/traits/testing/unittest_tools.py

contributes towards #120

### DIFF
--- a/apptools/selection/tests/test_list_selection.py
+++ b/apptools/selection/tests/test_list_selection.py
@@ -1,6 +1,6 @@
-import numpy
+import unittest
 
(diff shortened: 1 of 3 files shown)


kitchoi (approved):
LVGTM

PR #188, opened 2020-11-05
Remove mock compatibility imports on Python 2

rahulporuri:
This PR removes the compatibilty layer for `mock` imports for python 2 vs python 3 compatibility, now that the package is moving towards python 3 only. This PR also removes `mock` as a dependency from the `etstool` development utility.

contributes towards #120


kitchoi (approved):
LGTM

PR #189, opened 2020-11-05
Remove bare returns in preferences tests

kitchoi:
Bare returns at the end of a function is unnecessary. This PR removes the ones in `apptools.preferences.tests`.

This is a localized effort, so that when we try to add tests in the package, we don't have to be forced to add those returns because that's the consistent thing to do in the local context.
**Checklist**
- Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst): Not news worthy

### DIFF
--- a/apptools/preferences/tests/test_preference_binding.py
+++ b/apptools/preferences/tests/test_preference_binding.py
@@ -45,12 +45,8 @@ class PreferenceBindingTestCase(unittest.TestCase):
         # The filename of the example preferences file.
         self.example = resource_filename(PKG, 'example.ini')
 
-        return
-
     def tearDown(self):
         """ Called immediately after each test method has been called. """
-
-        return
 
     ###########################################################################
     # Tests.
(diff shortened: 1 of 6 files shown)


aaronayres35 (approved):
LGTM!

apptools/preferences/tests/test_preference_binding.py
```diff
@@ -45,13 +45,9 @@ def setUp(self):
         # The filename of the example preferences file.
         self.example = resource_filename(PKG, 'example.ini')
 
-        return
-
     def tearDown(self):
         """ Called immediately after each test method has been called. """
```
kitchoi, line 50: Hmmm, strange that this does not result in syntax error.

kitchoi: Ah it is because there is a docstring. OK

rahulporuri (approved):
LGTM

PR #190, opened 2020-11-05
Explicitly requires Python >= 3.6

kitchoi:
This PR updates `setup.py` and `etstool.py` to remove support for Python 2.7 and Python 3.5.

The changes on `setup.py` will be useful for package managers downstream that are automatically building open-source packages (I wish Python 2.7 was gone from the world but there exists systems that still requires it.)

Related to #120

**Checklist**
- [x] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst) <--- Yes, this is news worthy.

### DIFF
--- a/docs/releases/upcoming/190.removal.rst
+++ b/docs/releases/upcoming/190.removal.rst
@@ -1,0 +1,1 @@
+Remove support for Python 2.7 and 3.5 (#190)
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM

Issue #194, opened 2020-11-06
PreferencesHelper unable to synchronize mutations on nested containers


kitchoi:
Similar to #129, currently mutations on nested containers for traits defined on a subclass of PreferencesHelper cannot be synchronized with the Preferences.

e.g.
```
from apptools.preferences.api import Preferences, PreferencesHelper
from traits.api import Int, List, Str


class MyPreferencesHelper(PreferencesHelper):
    preferences_path = Str('my_section')
    list_of_list_of_str = List(List(Str))

pref = Preferences(filename="tmp.ini")
helper = MyPreferencesHelper(preferences=pref, list_of_list_of_str=[["1"]])
helper.list_of_list_of_str[0].append("9")
pref.save()

```

The saved preferences would be:
```
[my_section]
list_of_list_of_str = [['1']]
```

Expected:
```
[my_section]
list_of_list_of_str = [['1', '9']]
```

kitchoi:
Note that the example above with nested List of List did happen to work prior to Traits 6.1. More specifically, before this fix: https://github.com/enthought/traits/pull/1018
But if it was a list of dict or a list of set, it still wouldn't work.

PR #196, opened 2020-11-06
Fix container items change event being saved in preferences

### LINKED ISSUE #129: PreferencesHelper sets *_items preference for List preferences
Version info (although the same problem essentially exists with traits v6.0.0):
```bash
$ edm list | grep -P "apptools|traits "
apptools             4.5.0-5         enthought/free
traits               6.1.1-1         enthought/free
```

Here's a small example:

Initially, the preferences file contains:
```text
# File: myprefs.ini
[my_section]
things = "['one', 'two']"
```

Then, we run a script that updates the `things` preference list:

```python
from apptools.preferences.api import Preferences, PreferencesHelper
from traits.api import Int, List, Str

class MyHelper(PreferencesHelper):
    preferences_path = 'my_section'
    things = List(Str)

prefs = Preferences(filename='myprefs.ini')
helper = MyHelper(preferences=prefs)
helper.things.append('another thing!')
prefs.save()
```

After the script is run, the preferences file looks like:

```text
# File: myprefs.ini
[my_section]
things = "['one', 'two']"
things_items = "TraitListEvent(index=2, removed=[], added=['another thing!'])"
```

As far as I can tell, the culprit is this block in the PreferencesHelper class:

https://github.com/enthought/apptools/blob/5a136f642d6865970497aa3b2f1f176d56841de2/apptools/preferences/preferences_helper.py#L65-L73

kitchoi:
Closes #129

This PR fixes the particular situation in #129 where the TraitListEvent is saved in the preferences and that the updated list is not saved in the preferences. This should fix a similar scenario for a Set or Dict. However, this does not fix the situations for nested containers. see: https://github.com/enthought/apptools/issues/194. I have added a sort of 'expected-failure' test in this PR.

Reacting to changes in arbitrarily nested containers could be supported by observe. However, not knowing the structure of the traits being serialized in the preferences, one may require recursion support, which has not been implemented in Traits observe.

**Checklist**


apptools/preferences/preferences_helper.py
```diff
@@ -65,9 +72,23 @@ def _preferences_default(self):
...
+        # If the trait was a list or dict '_items' trait then just treat it as
+        # if the entire list or dict was changed.
+        if trait_name.endswith('_items'):
+            trait_name = trait_name[:-6]
+            if self._is_preference_trait(trait_name):
+                self.preferences.set(
+                    '%s.%s' % (self._get_path(), trait_name),
+                    getattr(self, trait_name)
+                )
+                return
```
kitchoi, line 87: This is ugly, and is similar to what the UI counterpart is doing:
https://github.com/enthought/apptools/blob/83aa787c33edc220562e448ad649450f50bdffa0/apptools/preferences/ui/preferences_page.py#L76-L81

The alternative might have been a redesign of the helper <-> preference relationship, but that might not be feasible without changing how it is used.

apptools/preferences/preferences_helper.py
```diff
@@ -65,9 +72,23 @@ def _preferences_default(self):
...
+
         # If we were the one that set the trait (because the underlying
         # preferences node changed) then do nothing.
-        if self.preferences and self._is_preference_trait(trait_name):
+        if self._is_preference_trait(trait_name):
```
aaronayres35, line 91: This is likely an unrealistic edge case (it would probably be considered a programmer error), but what if there is a preference trait that ends with `_items`, e.g. `something_items`?  In that case we enter the above if statement, and set `trait_name = = trait_name[:-6]` removing the trailing `_items` leaving us with `something`.  Then calling `getattr(self, trait_name)` will fail.  

In general you should never name traits with a trailing `_items`, so this code is good to stay as is. But just wanted to ask about this.
e.g. this test fails:
<details>

```
def test_mutate_list_of_values(self):
        """ Mutated list should be saved and _items events not to be
        saved in the preferences.
        """
        # Regression test for enthought/apptools#129

        class MyPreferencesHelper(PreferencesHelper):
            preferences_path = Str('my_section')

            list_of_str_items = List(Str)

        helper = MyPreferencesHelper(list_of_str_items=["1"])

        # Now modify the list to fire _items event
        helper.list_of_str_items.append("2")
        self.preferences.save(self.tmpfile)

        new_preferences = Preferences()
        new_preferences.load(self.tmpfile)

        self.assertEqual(
            new_preferences.get("my_section.list_of_str_items"), str(["1", "2"])
        )
        self.assertEqual(new_preferences.keys("my_section"), ["list_of_str_items"])

```

</details>

aaronayres35 (approved):
LGTM!

No suggested changes, just one observation / question

apptools/preferences/preferences_helper.py
kitchoi: Yep, it is a very good observation. I agree this is why magic naming is bad, and one of the reasons to introduce `observe` is to do away with `_items` magic naming.

kitchoi: I think the test you provided can be fixed though, but it may have to be done along with `apptools.preferences.ui.preferences_page`. I will do that in a separate PR.

To be honest, I think a better design would be one that does not require synchronization - we probably don't need that. Most of the time we care about storing the preferences to a file after it has been edited. As long as we can obtain and serialize all the values at the point when the file is saved, we are good. If someone says they want to share the same preferences with multiple helpers, I'd say, think again. A lot can go wrong when editing shared mutable states, best approach is don't.

kitchoi: > I think the test you provided can be fixed though, but it may have to be done along with apptools.preferences.ui.preferences_page. I will do that in a separate PR.

Actually thinking about it... I am not sure it can be fixed to satisify both situations. The suffix `_items` is simply overloaded. It might be one that just needs to be documented and a warning be emitted perhaps.

apptools/preferences/preferences_helper.py
```diff
@@ -65,9 +72,23 @@ def _preferences_default(self):
     def _anytrait_changed(self, trait_name, old, new):
         """ Static trait change handler. """
 
+        if self.preferences is None:
+            return
+
+        # If the trait was a list or dict '_items' trait then just treat it as
+        # if the entire list or dict was changed.
+        if trait_name.endswith('_items'):
+            trait_name = trait_name[:-6]
```
rahulporuri, line 81: when we get to python 3.9, we would be able to use `removesuffix` method - which would make the code more readable. ref https://www.python.org/dev/peps/pep-0616/

maybe we should add two backwards compatible `removeprefix` and `removesuffix` functions in a traits utility - which other libraries that depend on traits can use.

apptools/preferences/preferences_helper.py
kitchoi: Opened https://github.com/enthought/apptools/issues/225

PR #192, opened 2020-11-06
Remove u-string prefixes

rahulporuri:
The u-string prefixes arent needed on python 3

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ This PR is not news-worthy for end users.

PR #198, opened 2020-11-10
Refresh Preferences documentation

### LINKED ISSUE #42: update the documentation about Preferences
#41 has changed the way the Preferences are managed internally. They are no more only str. Documentation must be updated.

kitchoi:
Closes #42

This PR updates the Preferences section in the documentation.

I re-run the code mentioned in the documentation and updated the output.

I also removed a block which advertises mutating the class attribute (which has a trait type) directly. If it had worked before, it does not work any more. I am tempted to say this is probably not a good pattern anyway to mutate global state on PreferencesHelper so I did not open a new issue.

There are some minor, opportunistic visual updates in the sections/subsections organization.
For example, a couple of sections are at the same level as "Preferences" itself, so previously the TOC looks confusing:
Screenshot 2020-11-10 at 15 54 47

Now those sections are made subsections, so the TOC looks like this:
Screenshot 2020-11-10 at 16 20 27

**Checklist**

### DIFF
--- a/docs/source/preferences/Preferences.rst
+++ b/docs/source/preferences/Preferences.rst
@@ -8,16 +8,16 @@ mechanism and each layer on top providing more convenient ways to get and set
 preference values.
 
 The Basic Preferences Mechanism
-===============================
+-------------------------------
 
 Lets start by taking a look at the lowest layer which consists of the
-IPreferences_ interface and its default implementation in the Preferences_
+|IPreferences| interface and its default implementation in the |Preferences|
 class. This layer implements the basic preferences system which is a
 hierarchical arrangement of preferences 'nodes' (where each node is simply an
-object that implements the IPreferences_ interface). Nodes in the hierarchy can
+object that implements the |IPreferences| interface). Nodes in the hierarchy can
 contain preference settings and/or child nodes. This layer also provides a
 default way to read and write preferences from the filesystem using the
-excellent ConfigObj_ package.
+excellent `ConfigObj`_ package.
 
 This all sounds a bit complicated but, believe me, it isn't! To prove it
 (hopefully) lets look at an example. Say I have the following preferences in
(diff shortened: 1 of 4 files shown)


docs/source/preferences/Preferences.rst
```diff
@@ -162,7 +162,7 @@ preferences helper as follows::
   class SplashScreenPreferences(PreferencesHelper):
       """ A preferences helper for the splash screen. """
 
-      PREFERENCES_PATH = 'acme.ui'
+      preferences_path = 'acme.ui'
```
kitchoi, line 165: The use of `PREFERENCES_PATH` results in this warning message:
https://github.com/enthought/apptools/blob/4bf708f7b9e856c3fa4630c7de36821e31532ade/apptools/preferences/preferences_helper.py#L115

docs/source/preferences/Preferences.rst
```diff
@@ -202,30 +202,14 @@ preferences node directly::
   >>> preferences.set('acme.ui.ratio', 0.33)
   ratio 0.75 0.33
 
-If you always use the same preference node as the root of your preferences you
-can also set the class attribute 'PreferencesHelper.preferences' to be that
-node and from then on in, you don't have to pass a preferences collection in
-each time you create a helper::
-
-  >>> PreferencesHelper.preferences = Preferences(filename='example.ini')
```
kitchoi, line 210: This mutates class attribute (global state) and it also does not work... `helper.bgcolor` returns `""`, `helper.width` returns `0` (default of Int), etc.

docs/source/preferences/Preferences.rst
```diff
@@ -245,14 +229,14 @@ preferences is just like using the plain old non-scoped version::
...
-    Node() {}
-      Node(application) {}
-        Node(acme) {}
-          Node(ui) {'bgcolor': 'blue', 'ratio': '1.0', 'width': '50', 'visible': 'True'}
-            Node(splash_screen) {'image': 'splash', 'fgcolor': 'red'}
-      Node(default) {}
+   Node() {}
+     Node(application) {}
+       Node(acme) {}
+         Node(ui) {'bgcolor': 'blue', 'width': '50', 'ratio': '1.0', 'visible': 'True'}
```
kitchoi, line 237: Changed because since Python 3.7 dictionary order is preserved to match insertion order. So the output order is going to match the order in the preference file.

docs/source/preferences/PreferencesInEnvisage.rst
```diff
@@ -1,3 +1,5 @@
+.. _preferences-in-envisage:
+
```
kitchoi, line 2: Orthogonal... this section looks like it belongs to Envisage? Maybe a separate issue here.

kitchoi:
#42 is intended to document changes related to the difference between unicode and str, but since we are dropping Python 2, turns out there is no much we need to do regarding unicode. References to Unicode have been removed in #123 and #192

aaronayres35 (approved):
Changes LGTM!  I am yet to build the docs though

aaronayres35:
> Changes LGTM! I am yet to build the docs though

just built the docs and still LGTM

kitchoi:
Thanks. I will open the issue about the suspicious "Preferences in Envisage" section.

PR #200, opened 2020-11-10
remove dead code

aaronayres35:
Throughout the `apptools` code base there was a lot of commented out code.  Some of it is print statements or no longer used import statements.  Others are things like "we want to be able do this: (bunch of code)  but it doesn't work right now.  These things should be github issues and not clouding the code base.  If a reviewer finds anything deleted here to be important, it can be added back, or preferably have a github issue opened.  Other cases seem like they were part of a PR review process in the form of # actually we dont need this # some code.  These have been removed as well. 

**Checklist**

### DIFF
--- a/apptools/preferences/ui/i_preferences_page.py
+++ b/apptools/preferences/ui/i_preferences_page.py
@@ -21,3 +21,4 @@ class IPreferencesPage(Interface):
 
     def apply(self):
         """ Apply the page's preferences. """
+        pass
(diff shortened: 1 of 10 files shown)


apptools/preferences/ui/i_preferences_page.py
```diff
@@ -21,12 +21,3 @@ class IPreferencesPage(Interface):
 
     def apply(self):
         """ Apply the page's preferences. """
-
```
kitchoi, line 24: Part of me would want to see `pass` here (which is probably the more consistent thing to do?). I know I did not do that on another occasion when I removed the bare returns due to laziness. Just wanted to make sure I have not accidentally advocated a pattern when I did not mean it 😓 , sorry about that.

kitchoi (approved):
LGTM

Issue #57, opened 2015-10-01
Remove spickle entirely


prabhuramachandran:
The `apptools.persistence.spickle` module does not work on Python 3.4.  We either should port it so it does or remove it entirely if no one is using it.

rahulporuri:
I don't see any users of `apptools.persistence.spickle` so i think we can remove it. Ref https://github.com/search?p=1&q=%22apptools.persistence.spickle%22&type=Code

PR #201, opened 2020-11-10
Remove all checks for python version < 3 as we are dropping python2

aaronayres35:
related to #120 

Since #190 has been merged and we intend to drop python 2 I believe it is fine to remove these checks now.  The changes in this PR all occur in the `persistence` sub package which right now, AFAIK, we plan on keeping.

Additionally see here: 
https://github.com/enthought/apptools/blob/a693c79e8547ccaafcf89f230c0b17edbc780655/apptools/persistence/spickle.py#L19-L20

I left this for now as there is already a relevant issue open: #57 

**Checklist**

### DIFF
--- a/apptools/persistence/state_pickler.py
+++ b/apptools/persistence/state_pickler.py
@@ -129,7 +129,7 @@ def gunzip_string(data):
     """Given a gzipped string (`data`) this unzips the string and
     returns it.
     """
-    if (bytes is not str and type(data) is bytes):
+    if type(data) is bytes:
         s = BytesIO(data)
     else:
         s = StringIO(data)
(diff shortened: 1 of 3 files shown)


kitchoi (approved):
LGTM

(I am assuming this is not news-worthy for the end-users so there are no need for a news fragment)

apptools/persistence/state_pickler.py
```diff
@@ -130,7 +129,7 @@ def gunzip_string(data):
     """Given a gzipped string (`data`) this unzips the string and
     returns it.
     """
-    if PY_VER== 2 or (bytes is not str and type(data) is bytes):
+    if (bytes is not str and type(data) is bytes):
```
rahulporuri, line 132: Umm. `bytes is not str` should always return `True` in Python 3. We should be able to get rid of it.

Issue #145, opened 2020-09-28
Make codebase Flake8-clean


corranwebster:
Probably OK to simply run Black over it.

mdickinson:
> Probably OK to simply run Black over it.

Definitely fine to run `black` over the codebase (as a one-time thing), but note that that only solves a handful of flake8 issues.

PR #203, opened 2020-11-11
Setup flake8

rahulporuri:
This PR sets up flake8 for the codebase. A new `flake8` command is added to `etstool.py` but all of `apptools` is excluded when running `flake8`. The idea is to slowly get the codebase flake8-clean. Note that `examples`, `integrationtests` and `etstool.py`, `setup.py` are also flake8 targets.

contributes towards #145 

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ This has no effect on end users so no news fragment is added for this PR.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,19 @@
+[flake8]
+ignore = E266, W503
+exclude =
+    # The following list should eventually be empty.
+    apptools/_testing/*
+    apptools/help/*
+    apptools/io/*
+    apptools/logger/*
+    apptools/naming/*
+    apptools/persistence/*
+    apptools/preferences/*
+    apptools/scripting/*
+    apptools/selection/*
+    apptools/sweet_pickle/*
+    apptools/type_manager/*
+    apptools/type_registry/*
+    apptools/undo/*
+    examples/*
+    integrationtests/*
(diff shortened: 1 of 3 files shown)


aaronayres35 (approved):
LGTM!

setup.cfg
```diff
@@ -0,0 +1,21 @@
+[flake8]
+ignore = E266, W503
```
kitchoi, line 2: Are these two temporary or do they mean to stay (i.e. we really don't care about them)?
It'd be good to add a comment if the ignore should be empty too.

rahulporuri: Sorry for not pointing this out but yes, we don't care about them and those ignores are meant to stay. We ignore those errors in other ETS projects as well.

kitchoi: Thanks. I did not know that. Do you know if it is written somewhere? (it is not just some temporary thing ends up being thought of as permanent)

rahulporuri: :/ dont think its written anywhere afaik.

E266 - too many leading ‘#’ for block comment
W503 - line break before binary operator

https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

E266 errors have been observed because of our use of comments of the style

```
#### object interface #########
```

kitchoi: This is good to know. Thank you very much.

PR #204, opened 2020-11-11
Ensure Apptools works with toolkit name "qt" as well as "qt4"

### LINKED ISSUE #197: Ensure Apptools work with toolkit name "qt" as well as "qt4"
We have introduced "qt" as an alias for "qt4" as a toolkit name.  We should ensure that this works everywhere in apptools.  In a quick search of the current code, I think this is the only place where change is needed:

https://github.com/enthought/apptools/blob/ac27f663052f49c29e376c9377c70ad2c56a6909/examples/undo/example_editor_manager.py#L135

aaronayres35:
fixes #197 

This PR makes two changes to try and allow 'qt' to be accepted as a toolkit name as an alias for 'qt4'.  In one occasion (the one mentioned on the issue) I simply changed a check for `tk_name == 'qt4'` to  `tk_name == 'qt4' or tk_name == 'qt'`.

On another occasion we try to use `ETSCONFIG.toolkit` as part of a path for an import.  I believe using 'qt' here would break things because `traitsui.qt.editor` does not exist only `traitsui.qt4.editor` does.  Unfortunately I don't believe this will be exercised by the test suite, so this change was motivated just by my own suspicion.

**Checklist**


apptools/preferences/ui/widget_editor.py
```diff
@@ -7,7 +7,8 @@
 from traitsui.api import EditorFactory
 
 # fixme: We need to import the 'Editor' class from the appropriate toolkit.
-exec('from traitsui.%s.editor import Editor' % ETSConfig.toolkit)
+toolkit = {'wx': 'wx', 'qt4': 'qt4', 'qt': 'qt4'}[ETSConfig.toolkit]
+exec(f'from traitsui.{toolkit}.editor import Editor')
```
corranwebster, line 11: The right way to do this is to get the toolkit object from traitsui, and ask for the editor, something like:
```suggestion
from traitsui.toolkit import toolkit_object
Editor = toolkit_object('editor.Editor')
```

corranwebster, line 11: The right way to do this is to get the toolkit object from traitsui, and ask for the editor, something like:
```suggestion
from traitsui.toolkit import toolkit_object
Editor = toolkit_object('editor.Editor')
```

apptools/preferences/ui/widget_editor.py
```diff
@@ -7,7 +7,8 @@
 from traitsui.api import EditorFactory
 
 # fixme: We need to import the 'Editor' class from the appropriate toolkit.
```
rahulporuri, line 9:
```suggestion
```

rahulporuri (approved):
Mostly LGTM with one comment.

PR #206, opened 2020-11-12
Update flake8 ignores

kitchoi:
Drive-by updates to the flake8 settings.

- `apptools/_testing` is already flake8-clean on master.
- `docs/source/conf.py` should be flaked too? (I can revert that)
- We expect unused imports in `*/api.py` files, skip all of them (so we don't have to laboriously skip F401 at every single line). 

**Checklist**
- ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ Not news worthy

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,6 @@
 [flake8]
 ignore = E266, W503
+per-file-ignores = */api.py:F401
 exclude =
     # The following list should eventually be empty.
     apptools/_testing/*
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM

setup.cfg
```diff
@@ -15,6 +15,7 @@ exclude =
     apptools/type_manager/*
     apptools/type_registry/*
     apptools/undo/*
+    docs/source/conf.py
```
kitchoi, line 18: (I found this because I did not use etstool.py, I just used a Python venv with flake8 installed... so naturally I just did `flake8 .` and errors for this file turn up.)

PR #207, opened 2020-11-13
Replace pkg_resources with importlib_resources

rahulporuri:
This PR replaces the use of `pkg_resources.resource_filename` with `importlib_resources.files`. Ref the migration guide for `importlib_resources` - https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ This has no effect on the end users so this does not warrant a news fragment.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -103,6 +103,7 @@ dependencies = {
     "traitsui",
     "configobj",
     "coverage",
+    "importlib_resources",
     "pytables",
     "pandas",
     "pyface",
--- a/etstool.py
+++ b/etstool.py
@@ -103,7 +103,7 @@ dependencies = {
     "traitsui",
     "configobj",
     "coverage",
-    "importlib_resources",
+    "importlib_resources>=1.1.0",
     "pytables",
     "pandas",
     "pyface",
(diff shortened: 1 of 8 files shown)


etstool.py
```diff
@@ -103,6 +103,7 @@
     "traitsui",
     "configobj",
     "coverage",
+    "importlib_resources",
```
kitchoi, line 106: Does this have a version range requirement `importlib_resources >= 1.1.0`?

This is now a test dependency, but really, apptools has plenty of test dependencies that are not runtime dependencies and apptools has not been maintaining a separate set of dependencies in `setup.py`, but I think it should.  Ideally, we should be able to run the test suite using `pip`:
```
pip install apptools[test]
python -m unittest discover apptools
```

What do you think about adding a test dependency group in setup.py as part of this PR? (Just the `importlib_resources` one would be good enough for now).

rahulporuri: thanks for pointing that out. you're right, we need v `>=1.1.0`. I'll add a test dependency group.

kitchoi (approved):
LGTM

rahulporuri:
Appveyor CI MacOS job didn't start. at all. restarting it to see if it works this time around.

PR #215, opened 2020-11-16
Remove apptools.help subpackage and all references to it

### LINKED ISSUE #142: Decide fate of apptools.help
This is a workbench/envisage plugin for contributing links to external documentation and demos.  It uses preferences for pointing to directories which contain documentation and demo code.

It may make sense to keep for applications still using workbench; it may also make sense to create a Tasks version as well.

It may be used by Mayavi, but that is not clear.

aaronayres35:
closes #142 

This PR removes the `help` sub-package and any references to it in apptools.  Note that no other code in apptools was using it.  `logger` was trying to use `apptools.help.IHelp` but that did not exist.  The function referring to`apptools.help.IHelp` was actually never called anywhere else in `logger`, so that function was simply deleted.

**Checklist**


rahulporuri (approved):
LGTM

PR #216, opened 2020-11-16
Remove NullHandler from apptools.logger

### LINKED ISSUE #213: Remove apptools.logger.null_handler
The [`null_handler`](https://github.com/enthought/apptools/blob/master/apptools/logger/null_handler.py) module is now available from the standard library so it can be removed from `apptools.logger`.

aaronayres35:
fixes #213 

As mentioned on the issue it is now available from the standard library.  This PR simply deletes the `null_handler` module and removes the `NullHandler` import from `apptools.logger.api`

**Checklist**


rahulporuri (approved):
LGTM

PR #217, opened 2020-11-16
remove FilteringHandler and apptools.logger.util.py

### LINKED ISSUE #212: apptools.logger.filtering_handler.FilteringHandler can be removed
Offline discussion: `FilteringHandler` does not look like it would work as expected, and there are no apparent downstream usage.

This particular class can be removed, and with that `apptools.utils.get_module_name` and `apptools.utils.get_module_name_from_zip` can be removed.

aaronayres35:
fixes #212 

As described in the issue, `FilteringHandler` does not look like it would work as expected, and there are no apparent downstream usage.  Further, `FilteringHandler` was the only place where something from `apptools.logger.util.py` was called.  The majority of the code in `apptools.logger.util.py` was also labelled as WIP.  This PR deletes both the `apptools.logger.filtering_handler.py` and `apptools.logger.util.py` modules.  

**Checklist**


rahulporuri (approved):
LGTM

PR #221, opened 2020-11-18
Document apptools.naming

### LINKED ISSUE #139: Documentation for apptools naming
Apptools naming is complex and almost completely undocumented.  It is used by Mayavi, it cannot be ignored as not to be used, but without documentation it is not clear what the module is intended to be used for.

It looks like this may be a Python implementation of the [Java Naming and Directory Interface](https://en.wikipedia.org/wiki/Java_Naming_and_Directory_Interface#:~:text=The%20Java%20Naming%20and%20Directory,independent%20of%20the%20underlying%20implementation.).

aaronayres35:
Closes #139 (for now at least, we may want to bulk up the documentation at some point)

This PR adds a very small section to the apptools documentation for the `naming` subpackage.  All it says is that this is a python implementation of the Java Naming and Directory Interface, and provides a couple links to JDNI documentation. 

Note the two links are a little redundant (there is repeated information between them), but some things are different.  (I'm not exactly sure why both links exist, and the content isn't merged)

**Checklist**

### DIFF
--- a/docs/source/naming/Introduction.rst
+++ b/docs/source/naming/Introduction.rst
@@ -1,0 +1,6 @@
+Naming
+======
+
+Apptools naming is a python implementation of the naming portion of the `Java
+Naming and Directory Interface <https://docs.oracle.com/javase/tutorial/jndi/TOC.html>`_,
+including specific implementations for a heirarchy of Python objects.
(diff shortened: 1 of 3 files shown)


docs/source/naming/Introduction.rst
```diff
@@ -0,0 +1,7 @@
+Naming
+======
+
+Apptools naming is a python implementation of the naming portion of the `Java
```
kitchoi, line 4: We can provide cross link to the API reference. The word "naming" also gets merged into the sentence and it is hard to see it is actually a package name.
```suggestion
:mod:`apptools.naming` package is a Python implementation of the Naming portion of the `Java
```

kitchoi (approved):
LGTM. Minor nitpicks.

PR #226, opened 2020-11-19
Fix use of _items as a normal trait in PreferencesHelper and PreferencesPage

### LINKED ISSUE #225: PreferencesHelper does not distinguish defined trait named "*_items"  from _items event trait for listeners
In PreferencesPage, we have this block of code that synchronizes changes on the PreferencesPage with the underlying Preferences:
https://github.com/enthought/apptools/blob/d755bf155e0ef8a1e5d34fcfec50bd6666effe08/apptools/preferences/ui/preferences_page.py#L66-L85

#196 makes the PreferencesHelper (the base class of `PreferencesPage`) consistent in the handling of `_items` trait: It is assumed to be the event trait for listening to list/dict/set mutations.

This means one cannot have a trait that is actually named `*_items` and have it synchronized:
```
class MyPreferencesPage(PreferencesPage):
    name_items = Str()
```

If this cannot be supported, we should document it and perhaps emit a warning when the trait is skipped.

kitchoi:
Closes #225

This PR changes the listener for synchronizing the traits on the PreferencesHelper and PreferencesPage with the preferences, such that if the helper / page has a trait with a name suffix `_items` defined, it would not be mistaken as the event trait for list/set/dict mutation. Otherwise it would assume it is the event trait for list/set/dict mutation, and handle it specifically. If the assumption is wrong, the change handler will raise an error, but the error will be captured in a normal run condition and the default log message will appear.

This fix is possible because one cannot have a trait called `name` and a trait called `name_items` (not an event) and expect both of them to operate normally. Consider the following traits-only tests:
```
import unittest
from traits.api import *

class ItemsFirst(HasTraits):
    names_items = Str()
    names = List(Str)

class ItemsSecond(HasTraits):
    names = List(Str)
    names_items = Str()

class TestItems(unittest.TestCase):
    def test_mutate_fails_items_second(self):
        obj = ItemsSecond()
        obj.names_items = "Hello"
        obj.names = []
        with self.assertRaises(TraitError):
            obj.names.append("1")

    def test_assignment_fails(self):
        obj = ItemsFirst()
        obj.names = []
        obj.names.append("1")
        with self.assertRaises(TraitError):
            obj.names_items = "Hello"
```
The above tests pass. If there is a list called "names" and there is a trait called "names_items", the existence of the second trait will cause a lot of unexpected behaviours on its own already.

**Checklist**

### DIFF
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -75,21 +75,20 @@ class PreferencesHelper(HasTraits):
         if self.preferences is None:
             return
 
+        # If we were the one that set the trait (because the underlying
+        # preferences node changed) then do nothing.
+        if self._is_preference_trait(trait_name):
+            self.preferences.set("%s.%s" % (self._get_path(), trait_name), new)
+
         # If the trait was a list or dict '_items' trait then just treat it as
         # if the entire list or dict was changed.
-        if trait_name.endswith('_items'):
+        elif trait_name.endswith('_items'):
             trait_name = trait_name[:-6]
             if self._is_preference_trait(trait_name):
                 self.preferences.set(
                     '%s.%s' % (self._get_path(), trait_name),
                     getattr(self, trait_name)
                 )
-                return
-
-        # If we were the one that set the trait (because the underlying
-        # preferences node changed) then do nothing.
-        if self._is_preference_trait(trait_name):
-            self.preferences.set("%s.%s" % (self._get_path(), trait_name), new)
 
         return
 
@@ -205,4 +204,4 @@ class PreferencesHelper(HasTraits):
         ):
             return False
 
-        return True
+        return trait_name in self.editable_traits()
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -75,8 +75,6 @@ class PreferencesHelper(HasTraits):
         if self.preferences is None:
             return
 
-        # If we were the one that set the trait (because the underlying
-        # preferences node changed) then do nothing.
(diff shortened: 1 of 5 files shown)


aaronayres35 (approved):
Changes LGTM.
The test that sparked my concern here: https://github.com/enthought/apptools/pull/196/files#r526423040 now passes as well. (This scenario covered by the tests added here).  
Also this PR description is very clear as to why this fix is possible

apptools/preferences/preferences_helper.py
```diff
@@ -75,21 +75,20 @@ def _anytrait_changed(self, trait_name, old, new):
         if self.preferences is None:
             return
 
+        # If we were the one that set the trait (because the underlying
+        # preferences node changed) then do nothing.
+        if self._is_preference_trait(trait_name):
+            self.preferences.set("%s.%s" % (self._get_path(), trait_name), new)
```
aaronayres35, line 81: I do find this comment a little confusing... We are not conditionally doing nothing as it suggests

kitchoi: Good point. I think I know what this means, but I agree this is not obvious at all, I will try to fix this comment.

apptools/preferences/preferences_helper.py
```diff
@@ -90,6 +88,9 @@ def _anytrait_changed(self, trait_name, old, new):
                     getattr(self, trait_name)
                 )
 
+        # If the change refers to a trait defined on this class, then
+        # the trait is not a preference trait and we do nothing.
+
```
aaronayres35, line 93: This is much clearer, thank you!

PR #227, opened 2020-11-19
Some Flake8 updates

aaronayres35:
Part of #145
Some of the changes are also related to #120 

**Checklist**

### DIFF
--- a/apptools/io/file.py
+++ b/apptools/io/file.py
@@ -15,13 +15,18 @@
 
 
 # Standard/built-in imports.
-import mimetypes, os, shutil, stat
+from functools import total_ordering
+import mimetypes
+import os
+import shutil
+import stat
 
 # Enthought library imports.
 from traits.api import Bool, HasPrivateTraits, Instance, List, Property
 from traits.api import Str
 
 
+@total_ordering
 class File(HasPrivateTraits):
     """ A representation of files and folders in a file system. """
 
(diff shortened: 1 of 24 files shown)


apptools/io/file.py
```diff
@@ -82,12 +87,23 @@ def __init__(self, path, **traits):
...
+        if isinstance(other, File):
+            return not (self == other)
+
+        return True
+
+    def __lt__(self, other):
+        if isinstance(other, File):
+            return self.path < other.path
+
+        return False
```
kitchoi, line 106: I believe this `File` has been used in projects using Python 3, and the `__cmp__` is simply not honoured any more. Implementing `__eq__` here will be introducing a new behaviour for projects using `apptools.io.File` in Python 3. 

Shall we simply remove the `__cmp__` here and open a separate issue as to whether to reimplement the comparison behaviour?  It is possible that these comparisons are not needed. `__eq__` and comparisons are used in places like `if x in list_of_files` and `sorted(list_of_files)`. It is hard to revert because they are used implicitly so I'd generally defer implementing them until I am sure the logic is generally applicable and useful.

Other thoughts for this separate issue:
(The existing logic of comparing the file using the path string as is a bit questionable...what if the path is a Window flavour-ed one and the other one has been normalized to Unix path?)
(For the `__lt__`, it probably should return `NotImplemented` if the object is not an instance of `File`)
(For Python 3, we don't need to implement `__ne__` any more if there already exists an `__eq__`. The opposite is true for Python 2.)

kitchoi (commented):
One comment on the comparison behaviour for `apptools.io.File`. The rest looks good to me.

kitchoi (approved):
LGTM

PR #229, opened 2020-11-20
More Flake8 cleanup

aaronayres35:
part of #145 
With this PR the codebase is very nearly fully flake8 clean.  There is the possibility some of the remaining non-flake8 clean code will be killed as well.

**Checklist**


docs/source/conf.py
```diff
@@ -17,13 +18,13 @@
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
 # absolute, like shown here.
-#sys.path.append(os.path.abspath('some/directory'))
+# sys.path.append(os.path.abspath('some/directory'))
```
aaronayres35, line 21: I should probably just delete all these commented out code blocks

aaronayres35: Actually I'm not sure, some of it isn't really dead code, but rather unused options.  I removed all of it in the latest commit, but if these should be kept I can simply revert that commit

integrationtests/persistence/test_persistence.py
```diff
@@ -17,11 +17,11 @@ def __init__(self, firstname, lastname):
 class Foo:
```
aaronayres35, line 17: TBH these integration tests could probably just be deleted.  I don't believe they ever get run/can be run

apptools/scripting/recorder.py
```diff
@@ -404,10 +403,7 @@ def save(self, file):
         """Save the recorded lines to the given file.  It does not close
         the file.
         """
-        if six.PY3:
-            file.write(self.get_code())
-        else:
-            file.write(six.text_type(self.get_code(), encoding="utf-8"))
```
aaronayres35, line 410: This change is not a flake8 cleaning, it is just related to dropping pyhton 2 support #120 
I added it because the later case below with `six.PY2` had a flake8 error, so I removed it, and it felt appropriate to remove them together.

aaronayres35: This sparked my opening of #230 which makes the same change, but goes further to replace all uses of `six` throughout the code base.

integrationtests/persistence/test_persistence.py
kitchoi: There might be values in converting them to test cases so that they can be discovered and run? In any case, that would be a separate issue. Agreed that tests don't get automatically run typically get forgotten and end up bring broken long before anyone realizes.

kitchoi (approved):
LGTM

integrationtests/persistence/test_persistence.py
aaronayres35: Yeah that makes sense.  I will leave that for a separate issue/PR

PR #230, opened 2020-11-20
Remove all use of six 

aaronayres35:
part of #120 

This PR removes all uses of `six` and replaces the occurrences with just the python 3 specific code. 

**Checklist**
Not adding a news fragment as one already exists saying "Remove support for Python 2.7 and 3.5".  That is what is relevant to `apptools` users.

### DIFF
--- a/apptools/type_registry/type_registry.py
+++ b/apptools/type_registry/type_registry.py
@@ -56,7 +56,7 @@ class TypeRegistry(object):
         obj : object
             The object to register.
         """
-        if isinstance(typ, six.string_types):
+        if isinstance(typ, str):
             # Check the cached types.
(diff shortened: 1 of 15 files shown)


apptools/type_registry/type_registry.py
```diff
@@ -56,7 +56,7 @@ def push(self, typ, obj):
         obj : object
             The object to register.
         """
-        if isinstance(typ, six.string_types):
+        if isinstance(typ, str):
```
kitchoi, line 59: 🤖  `six` is imported but unused in this file (on line 1).

kitchoi (approved):
LGTM.

Issue #146, opened 2020-09-28
Decide fate of apptools.naming


corranwebster:
This is an implementation of the [Java Naming and Directory Interface](https://en.wikipedia.org/wiki/Java_Naming_and_Directory_Interface) - or at least the naming part of that - with specific implementations for a hierarchy of Python objects and for file system stores.  It is used by Mayavi, but possibly only as part of the Wx backend.

It is intended to provide a standard API for hierarchical namespaces that standard interfaces can then build on top of, but I have a suspicion that at least some of the code that duplicates the Java APIs is over-engineered for Python (eg. the `InitialContextFactory` object feels very much like it could just be a callable).

If it is to be kept, then there are some improvements that should be made:

- it has its own adapter system; this should be converted to use standard Traits adaptation
- the pyfs_context seems of limited utility for Enthought and insecure by default (since it uses pickles); a JSON-backed version might make more sense if it is needed at all
- perhaps split out implementations from abstract/base classes
- separate out the UI components and fix them up (see #138)
- generally clean-up the codebase.

aaronayres35:
Coverage report on `naming` from running the full test suite:

<img width="1051" alt="Screen Shot 2020-11-10 at 11 21 44 AM" src="https://user-images.githubusercontent.com/36972686/98708586-f2048f00-2346-11eb-843f-a17d9c365f3b.png">

<img width="1086" alt="Screen Shot 2020-11-10 at 11 19 22 AM" src="https://user-images.githubusercontent.com/36972686/98708313-a3ef8b80-2346-11eb-9584-2e66dfef1e39.png">

<img width="1069" alt="Screen Shot 2020-11-10 at 11 20 09 AM" src="https://user-images.githubusercontent.com/36972686/98708394-b669c500-2346-11eb-87ff-2030d2de2199.png">

kitchoi:
Offline discussion: `apptools.naming` will be kept (at least for the next release: 5.0) as it is being used downstream (e.g. mayavi).

kitchoi:
> it has its own adapter system; this should be converted to use standard Traits adaptation

That adapter system is removed as part of #219 

> Offline discussion: apptools.naming will be kept (at least for the next release: 5.0) as it is being used downstream (e.g. mayavi).

I think when we drew this conclusion, we had not explored in details the functionality offered by `apptools.naming.ui.api.explore`. 

Through #219, we now have a better understanding as to what `apptools.naming.ui.api.explore` does.

[...]

PR #233, opened 2020-11-23
Remove apptools.naming.ui submodule

aaronayres35:
As mentioned here: https://github.com/enthought/apptools/issues/146#issuecomment-732281279, it has been decided that we will remove the `apptools.naming.ui` submodule.  This PR simple deletes the submodule and any reference to it.

Note I deleted images from the `LICENSE.txt` file as the modules containing them have now been deleted.  It seems to me that should be okay to do.

**Checklist**


image_LICENSE.txt
```diff
@@ -12,10 +12,3 @@ Unless stated in this file, icons are work of Enthought, and are released
 under BSD-like license.
 Files and original authors:
 ----------------------------------------------------------------
```
rahulporuri, line 14:
```suggestion

```

rahulporuri: and you can remove `GV: ...` from the earlier `Project/License/File` list.

rahulporuri: this can be done in a separate PR but it looks like the image license file needs to be updated. After this PR, it looks like there will be no images left in `apptools`. can you open a separate issue for this please?

rahulporuri (commented):
Mostly LGTM. There are a couple of comments and there is a naming/explorer example that will also need to be removed.

image_LICENSE.txt
aaronayres35: opened #236

rahulporuri (approved):
LGTM

PR #234, opened 2020-11-23
Remove occurrences of `from __future__ import print_function`

aaronayres35:
Part of #120 

This PR simply removes occurrences of `from __future__ import print_function` across the `apptools` codebase.

**Checklist**


rahulporuri (approved):
LGTM

PR #239, opened 2020-11-25
Final flake8 fixes

### LINKED ISSUE #145: Make codebase Flake8-clean

aaronayres35:
fixes #145 

After this PR, the `setup.cfg` exclude section is entirely empty.  We still have the per-file-ignores for F401 errors in `api` and `__init__` files which should stay.  We also still ignore E266 and W503 errors but I believe that is intended to remain the case.  

**Checklist**

### DIFF
--- a/apptools/naming/dynamic_context.py
+++ b/apptools/naming/dynamic_context.py
@@ -82,7 +82,7 @@ class DynamicContext(Context):
         item = self._get_contributed_context_item(name)
         if item != (None, None):
             obj, trait = item
-            result = True == trait.is_context
+            result = trait.is_context is True
         else:
             result = False
 
@@ -168,7 +168,7 @@ class DynamicContext(Context):
         traits = [(t.context_order, n, t) for n, t in traits.items()]
         traits.sort()
 
-        # Convert these trait definitions into a list of name and object tuples.
+        # Convert these trait definitions into a list of name and object tuples
         result = [
             (t.context_name, getattr(self, n), t) for order, n, t in traits
         ]
--- a/apptools/naming/dynamic_context.py
+++ b/apptools/naming/dynamic_context.py
@@ -82,7 +82,7 @@ class DynamicContext(Context):
         item = self._get_contributed_context_item(name)
         if item != (None, None):
             obj, trait = item
-            result = True if trait.is_context is True else False
+            result = trait.is_context is True
         else:
             result = False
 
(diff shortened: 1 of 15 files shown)


apptools/naming/dynamic_context.py
```diff
@@ -82,7 +82,7 @@ def _is_context(self, name):
         item = self._get_contributed_context_item(name)
         if item != (None, None):
             obj, trait = item
-            result = True == trait.is_context
+            result = trait.is_context is True
```
rahulporuri, line 85:
```suggestion
            result = True if trait.is_context is True else False
```

apptools/naming/pyfs_context.py
```diff
@@ -232,7 +231,7 @@ def _lookup(self, name):
                     # If the load fails then we create a generic file resource
                     # (the idea being that it might be useful to have access to
                     # the file to see what went wrong).
-                    except:
+                    except Exception:
```
rahulporuri, line 234: I would personally prefer leaving a `# noqa: EXXX` here instead of adding a bare `Exception`. That way, someone searching for flake8 ignores in the futures will notice this and maybe find time to investigate - instead of assuming that this is the behavior originally intended.

apptools/naming/trait_defs/naming_traits.py
```diff
@@ -73,7 +71,7 @@ def validate(self, object, name, value):
                     self.validate_failed(object, name, value)
             try:
                 value = self._get_binding_for(value)
-            except:
+            except Exception:
```
rahulporuri, line 74: same comment here about the use of `# noqa: EXXX` instead of adding a bare `Exception`

rahulporuri (commented):
Mostly LGTM

apptools/naming/dynamic_context.py
aaronayres35: I didn't spend much time looking into this but im quite confused by this code.  Why exactly can't we just use `result = traits.is_context`?  I'm assuming it is because `traits.is_context` could take other values besides true or false

rahulporuri: >  I'm assuming it is because traits.is_context could take other values besides true or false

that's what i assume. not sure what `trait.is_context` is.

apptools/naming/dynamic_context.py
```diff
@@ -82,7 +82,7 @@ def _is_context(self, name):
         item = self._get_contributed_context_item(name)
         if item != (None, None):
             obj, trait = item
-            result = True == trait.is_context
+            result = True if trait.is_context is True else False
```
rahulporuri, line 85:
```suggestion
            result = trait.is_context is True
```

rahulporuri (approved):
LGTM except for one minor cleanup.

PR #240, opened 2020-11-25
Update and check Copyright headers

### LINKED ISSUE #193: Update and check copyright headers
- [ ] The source code copyright headers need to be updated
- [ ] The `flake8-ets` package needs to be installed into the environment so that the headers are checked and kept up-to-date.

aaronayres35:
fixes #193 

This PR adds `flake8_ets` as a dependency so that copyright headers are checked by flake8, and also updates the copyright headers across apptools.

The only ones which may still need updating (they will currently cause flake8 errors which will cause the CI to fail on this PR) are some of those coming from `apptools.undo` which mention "Riverbank Computing Limited", e.g:
```
# ------------------------------------------------------------------------------
# Copyright (c) 2008, Riverbank Computing Limited
# All rights reserved.
#
# This software is provided without warranty under the terms of the BSD
# license included in enthought/LICENSE.txt and may be redistributed only
# under the conditions described in the aforementioned license.  The license
# is also available online at http://www.enthought.com/licenses/BSD.txt
# Thanks for using Enthought open source!
#
# Author: Riverbank Computing Limited
# Description: <Enthought undo package component>
# ------------------------------------------------------------------------------
```
I'm not sure what the correct way to handle these cases is. Maybe these should just be ignored in `setup.cfg`? (that would make the most sense to me)

**Checklist**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,4 +1,10 @@
 [flake8]
 ignore = E266, W503
-per-file-ignores = */undo/*:H101, */api.py:F401, */__init__.py:F401,
-    apptools/undo/api.py:F401,H101, apptools/undo/action/api.py:F401,H101
+per-file-ignores = 
+    */api.py:F401,
+    */__init__.py:F401,
+    # ignore undo related errors because it has been copied to pyface and will
+    # soon be deprecated.  See enthought/apptools#243
+    */undo/*:H101,
+    apptools/undo/api.py:F401,H101,
+    apptools/undo/action/api.py:F401,H101
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,4 @@
 [flake8]
 ignore = E266, W503
-per-file-ignores = */api.py:F401, */__init__.py:F401
+per-file-ignores = */undo/*:H101, */api.py:F401, */__init__.py:F401,
+    apptools/undo/api.py:F401,H101, apptools/undo/action/api.py:F401,H101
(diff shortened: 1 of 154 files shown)


rahulporuri:
> Maybe these should just be ignored in setup.cfg? (that would make the most sense to me)

The "right thing to do" - add the enthought copyright header on top of the riverbank computing ltd copyright header.
The "quick thing to do which works for us" - ignore copyright header flake8 errors in `apptools.undo`, especially because we want to move it from apptools to pyface.

aaronayres35:
> > Maybe these should just be ignored in setup.cfg? (that would make the most sense to me)
> 
> The "right thing to do" - add the enthought copyright header on top of the riverbank computing ltd copyright header.
> The "quick thing to do which works for us" - ignore copyright header flake8 errors in `apptools.undo`, especially because we want to move it from apptools to pyface.

That makes sense, I will ignore them here.  Once we have `flake8-ets` set up on pyface we can add them in pyface.

setup.cfg
```diff
@@ -1,3 +1,4 @@
 [flake8]
 ignore = E266, W503
-per-file-ignores = */api.py:F401, */__init__.py:F401
+per-file-ignores = */undo/*:H101, */api.py:F401, */__init__.py:F401,
```
kitchoi, line 3: Could we split it up into multiple lines and comment why `apptools.undo` is ignored?
If it can reference the issue that mentions deprecating `apptools.undo` that would be idea.

It has occurred in the past that _someone_ would go and assume things are temporarily ignored and wasted time linting something that is about to die anyway.

aaronayres35: That makes sense.  I don't know that there is an issue open currently, but I will open one and mention it here

kitchoi: Thank you!

kitchoi (approved):
LGTM

PR #242, opened 2020-11-27
Remove empty returns

### LINKED ISSUE #241: Remove spurious returns at the end of functions and methods
The coding style at Enthought a long time ago recommended that every function have a `return` at the end, even if it returned nothing.

There are still a few examples of this pattern in the Apptools codebase.  They should be removed as part of the clean-up for a 5.0 release.

rahulporuri:
This PR removes empty returns in the codebase. fixes #241 

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ Not news worthy to end users.

### DIFF
--- a/apptools/naming/tests/test_pyfs_context.py
+++ b/apptools/naming/tests/test_pyfs_context.py
@@ -72,7 +72,6 @@ class PyFSContextTestCase(unittest.TestCase):
         context = PyFSContext(path="data")
         self.assertEqual(len(context.list_bindings("")), 3)
 
-
     def test_initialization_with_empty_environment(self):
         """ initialization with empty environmentt """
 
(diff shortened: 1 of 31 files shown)


apptools/naming/tests/test_pyfs_context.py
```diff
@@ -293,15 +275,11 @@ def test_destroy_subcontext(self):
         # Try to destroy it.
         self.assertRaises(NotContextError, context.destroy_subcontext, "sub/a")
 
-        return
```
kitchoi, line 296: This is an early termination of the test, which means the subsequent `assertRaises` has not been run in the past, and is now being run. Luckily, it still passes.

kitchoi (approved):
LGTM.
The PR description makes this sound easier than it actually is: Only the `return` that are at the end of a function can be removed without changing the control flow. Another exception is if the function API actually meant to return `None`. 

I expanded the code surrounding the changes to ensure these returns are indeed redundant... except one. There is one change to a test, but I think given the assertions pass, that's okay.

apptools/naming/tests/test_pyfs_context.py
rahulporuri: yeah. i was looking out for this when i first pushed the changes.

PR #245, opened 2020-12-07
Remove a redundant test

kitchoi:
This PR removes a test that demonstrates the (unexpected) behaviour noted in #194.

The test actually fails with Traits 6.0 because the particular setup of List of List happened to work there:
```
FAIL: test_nested_container_mutation_not_supported (apptools.preferences.tests.test_preferences_helper.PreferencesHelperTestCase)
Known limitation: mutation on nested containers are not
  File "/Users/kchoi/Work/ETS/apptools/apptools/preferences/tests/test_preferences_helper.py", line 316, in test_nested_container_mutation_not_supported
    self.assertEqual(
AssertionError: AssertionError not raised
```

There is no much apptools can do to fix that. The test is closer to testing the limitations of Traits rather than testing Apptools itself, hence it should be removed. 

**Checklist**
- ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~: The removed test had not been released in the past.

### DIFF
--- a/apptools/preferences/tests/test_preferences_helper.py
+++ b/apptools/preferences/tests/test_preferences_helper.py
@@ -288,36 +288,6 @@ class PreferencesHelperTestCase(unittest.TestCase):
         )
         self.assertEqual(new_preferences.keys("my_section"), ["list_of_str"])
 
-    def test_nested_container_mutation_not_supported(self):
-        """ Known limitation: mutation on nested containers are not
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM! 
Maybe we will want to re-add such a test when (if?) we drop compatibility with traits 6.0

PR #246, opened 2020-12-09
Maint: Remove found news fragments rather than indiscriminately deleting all fragments in the folder

kitchoi:
This PR makes a change to the `python etstool.py changelog build` command such that only the found news fragment files are removed.

Motivation: If a news fragment file was misnamed, the build command won't collect its content. If the build command deletes the entire directory, such an omission will be harder to spot. By deleting only the found files, one can verify that the folder containing news fragments is indeed empty after build (with clean).

**Checklist**
- ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ No change to developer workflow

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -387,6 +387,9 @@ def build_changelog(ctx):
     # still require some tweaking.
     contents = []
 
+    # Collect news fragment files as we go, and then optionally remove them.
+    handled_file_paths = []
+
     for type_, description in ctx.obj["type_to_description"].items():
         pattern = os.path.join(NEWS_FRAGMENT_DIR, f"*.{type_}.rst")
         file_paths = sorted(glob.glob(pattern))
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM!  I tested the command locally with a broken news fragment and all but that one were deleted as expected.

My only suggestion (feel free to ignore / defer to a later PR) is that it might be useful to warn or print something to the console if the directory is not empty after `python etstool.py changelog build` is run.  As you mention in the description: "If the build command deletes the entire directory, such an omission will be harder to spot. By deleting only the found files, one can verify that the folder containing news fragments is indeed empty after build (with clean)."  It would make it almost impossible to miss if even further if there are any omissions something is printed to the console

kitchoi:
Thank you.

> My only suggestion (feel free to ignore / defer to a later PR) is that it might be useful to warn or print something to the console if the directory is not empty after python etstool.py changelog build is run. As you mention in the description: "If the build command deletes the entire directory, such an omission will be harder to spot. By deleting only the found files, one can verify that the folder containing news fragments is indeed empty after build (with clean)." It would make it almost impossible to miss if even further if there are any omissions something is printed to the console

Sure thing. I will make this enhancement in a separate PR.

PR #244, opened 2020-12-07
Fix deprecation warning from tests using TraitPrefixMap

kitchoi:
Closes #223

**Checklist**
- ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ Only concern warnings from tests, not news worthy

PR #247, opened 2020-12-09
Remove use of TraitPrefixMap in documentation example

kitchoi:
This PR updates documentation to remove usage of `TraitPrefixMap` which has been deprecated since Traits 6.1.

The example is the same as the one #244 modified in tests. Since this is for documentation, I think we can assume latest Traits version and there are no need to rewrite a compatibility layer there.

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ (edited: Too minor)

### DIFF
--- a/docs/releases/upcoming/247.doc.rst
+++ b/docs/releases/upcoming/247.doc.rst
@@ -1,0 +1,1 @@
+Replace TraitPrefixMap with PrefixMap in documentation example (#247)
--- a/docs/releases/upcoming/247.doc.rst
+++ b/docs/releases/upcoming/247.doc.rst
@@ -1,1 +1,0 @@
-Replace TraitPrefixMap with PrefixMap in documentation example (#247)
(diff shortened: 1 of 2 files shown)


docs/releases/upcoming/247.doc.rst
```diff
@@ -0,0 +1 @@
+Replace TraitPrefixMap with PrefixMap in documentation example (#247)
```
kitchoi, line 1: I am in two minds whether to skip this as it is quite minor.

aaronayres35 (approved):
LGTM

docs/releases/upcoming/247.doc.rst
aaronayres35: The news fragment wasn't included in #244, I think I'm of the view it is probably unneeded here

It is perfectly fine with me if you wish to keep it though

kitchoi: Yep, I will remove it.

Issue #202, opened 2020-11-11
Move "Preferences in Envisage" documentation to Envisage?


kitchoi:
This documentation section "Preferences in Envisage" looks like it belong to Envisage...?:
https://github.com/enthought/apptools/blob/addd26bac39b299c47630485e58c5649348e5a00/docs/source/preferences/PreferencesInEnvisage.rst

Envisage has a similar section: https://github.com/enthought/envisage/blob/3a7ab1c293107462e2022862ef8d3b71a20156b6/docs/source/envisage_core_documentation/preferences.rst

corranwebster:
Yes, those pages can probably be merged in the Envisage docs.

aaronayres35:
We will likely want to copy the relevant portions of the docs in `apptools` to `envisage` before the upcoming release.  See comment: https://github.com/enthought/apptools/pull/248#discussion_r540107003

PR #248, opened 2020-12-09
fix a few broken links in documentation

### LINKED ISSUE #171: Numerous broken links in docs
Many of the links in the documentation are broken and clicking them leads to 404 file not found / other errors.  For example, the first few links on this page:
https://github.com/enthought/apptools/blob/master/docs/source/preferences/Preferences.rst

Some of these links are in docs that will be dropped as the corresponding submodule is planned to be removed.  However, any docs that we expect to keep should be updated and have links fixed.

aaronayres35:
fixes #171 

Note when writing issue 171 I noticed the broken links on https://docs.enthought.com/apptools/.  I had not built the documentation from master, and after doing so it seems that the majority of the links were working on master.  I went through and checked links from the docs built on master and those that were still broken I have fixed here (only a few).

**Checklist**

### DIFF
--- a/docs/source/preferences/PreferencesInEnvisage.rst
+++ b/docs/source/preferences/PreferencesInEnvisage.rst
@@ -7,7 +7,7 @@ This section discusses how an Envisage application uses the preferences
 mechanism. Envisage tries not to dictate too much, and so this describes the
 default behaviour, but you are free to override it as desired.
 
-Envisage uses the default implementation of the ScopedPreferences_ class which
+Envisage uses the default implementation of the |ScopedPreferences| class which
 is made available via the application's 'preferences' trait::
 
   >>> application = Application(id='myapplication')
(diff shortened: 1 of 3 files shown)


docs/source/preferences/PreferencesInEnvisage.rst
```diff
@@ -45,5 +45,8 @@ e.g. To contribute a preference file for my plugin I might use::
       def get_preferences(self, application):
           return ['pkgfile://mypackage:preferences.ini']
 
-.. _PreferencesHelper: ../../enthought/preferences/preferences_helper.py
-.. _ScopedPreferences: ../../enthought/preferences/scoped_preferences.py
+..
+   # substitutions
+
+.. |PreferencesHelper| replace:: :class:`~apptools.preferences.preferences_helper.PreferencesHelper`
+.. |ScopedPreferences| replace:: :class:`~apptools.preferences.scoped_preferences.ScopedPreferences`
```
aaronayres35, line 52: Note that there is an open issue to move "Preferences in Envisage" to `envisage` (#202)
I'm unsure if we want to do that before the upcoming release? (if this were removed/moved over it wouldn't be available until another envisage release happens)

rahulporuri: It would definitely be nice to copy the docs before the release and like you said, we definitely can't remove the docs from apptools entirely. We'll need to go through a couple of minor versions of apptools before we can finally remove the docs from apptools entirely and simply point to envisage docs.

kitchoi (approved):
LGTM.

PR #252, opened 2020-12-10
add news fragment for apptools.undo deprecation

aaronayres35:
PR #250 was merged without a news fragment, but it needs one.  This PR simply adds that news fragment.

**Checklist**


docs/releases/upcoming/250.deprecation.rst
```diff
@@ -0,0 +1 @@
+Deprecatee apptools.undo subpackage (undo was moved to pyface) (#250)
```
kitchoi, line 1:
```suggestion
Deprecate apptools.undo subpackage (undo was moved to pyface) (#250)
```

kitchoi (approved):
Good point. I overlooked it too.

PR #257, opened 2020-12-10
Add extras_require to setup.py for optional dependencies

### LINKED ISSUE #176: Add extra_requires in setup.py for optional dependencies
Several subpackages in apptools have external dependencies, e.g. `apptools.io.h5` requires NumPy and PyTables (and their dependencies). These are currently treated as optional dependencies, but they are not listed in the setup.py in anyway. 

These optional dependencies should be included in extra_requires in setup.py so that developers can install apptools with these opt-in features. The mapping should be documented, e.g. as an installation instruction in README.

Related to #99, #70 (see PR #86)

aaronayres35:
fixes #176 

This PR adds to the `extras_require` section in `setup.py` to include additional optional dependencies.  It also mentions these in the readme so they are more apparent (Should I add instructions saying to run `pip install apptools[h5]` for examples?  I thought that might be overkill, but was unsure).

Note in this PR when looking for any optional dependencies I saw `io` imports listed as third party imports, so I moved those to be with standard library imports 

**Checklist**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,8 +44,9 @@ The `apptools.preferences` package requires:
 
 * `configobj <http://pypi.python.org/pypi/configobj>`_
 
-To install with `apptools.preferences` dependencies:
-`$ pip install apptools[preferences]`
+To install with `apptools.preferences` dependencies::
+
+    `$ pip install apptools[preferences]`
 
 The `apptools.io.h5` package requires:
 
@@ -53,15 +54,21 @@ The `apptools.io.h5` package requires:
 * `pandas <https://pypi.org/project/pandas/>`_
 * `tables <https://pypi.org/project/tables/>`_
 
-To install with `apptools.io.h5` dependencies:
-`$ pip install apptools[h5]`
+To install with `apptools.io.h5` dependencies::
+
+    `$ pip install apptools[h5]`
 
 The `apptools.persistence` package requires:
 
 * `numpy <https://pypi.org/project/numpy/>`_
 
-To install with `apptools.persistence` dependencies:
-`$ pip install apptools[persistence]`
+To install with `apptools.persistence` dependencies::
+
+    `$ pip install apptools[persistence]`
+
+To install with additional test dependencies::
+
+    $ pip install apptools[test]
 
 Many of the packages provide optional user interfaces using Pyface and
 Traitsui. In additon, many of the packages are designed to work with the
--- a/README.rst
+++ b/README.rst
@@ -44,31 +44,15 @@ The `apptools.preferences` package requires:
 
 * `configobj <http://pypi.python.org/pypi/configobj>`_
 
-To install with `apptools.preferences` dependencies::
-
-    $ pip install apptools[preferences]
-
 The `apptools.io.h5` package requires:
 
 * `numpy <https://pypi.org/project/numpy/>`_
 * `pandas <https://pypi.org/project/pandas/>`_
 * `tables <https://pypi.org/project/tables/>`_
 
-To install with `apptools.io.h5` dependencies::
-
-    $ pip install apptools[h5]
-
 The `apptools.persistence` package requires:
 
 * `numpy <https://pypi.org/project/numpy/>`_
-
-To install with `apptools.persistence` dependencies::
-
-    $ pip install apptools[persistence]
-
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -310,6 +310,14 @@ def get_long_description():
...
                 "importlib-resources>=1.1.0",
             ],
+            "h5": [
+                "numpy",
+                "pandas",
+                "tables",
+            ],
+            "persistence": [
+                "numpy",
+            ],
```
aaronayres35, line 320: `configobj` is only used by preferences.  Should it be moved here as a `preferences` specific optional dependency?

kitchoi: While I agree with the move, we can't move this abruptly because the installation process for anyone depending on preferences will have to change. 

e.g. Mayavi will be adversely affected by such a change, see its install requirement:
https://github.com/enthought/mayavi/blob/61ef7c0eebbdca474b7cfdf6082bfc17fb5aaf59/mayavi/__init__.py#L10-L19

Currently `pip install mayavi` will bring in `apptools`, which will bring in `configobj`. Mayavi does depend on the preferences module. If we are to move `configobj` from install requirement to optional extras requirement, we need to warn about it much in advanced, and make the necessarily adjustment downstream. 

I think we can add `"preferences": ["configobj"]` here, without removing the one in `install_requires`. This is probably the first step to that migration. e.g. Mayavi will change its install requires from `"apptools"` to  `"apptools[preferences]"`

kitchoi (commented):
>  (Should I add instructions saying to run pip install apptools[h5] for examples? I thought that might be overkill, but was unsure)

Not an overkill, I actually think it is good to provide straightforward installation instructions that end users can copy and paste, instead of having to parse the package links and come up with the command themselves.
e.g. https://github.com/enthought/pyface#installation

(I am guessing this wasn't done in the past because `pip` wasn't as feature rich then?)

kitchoi:
And this probably deserves a news fragment :) 
Looks like we need a new category for this too, maybe `build`?

kitchoi:
> Looks like we need a new category for this too, maybe build?

See https://github.com/enthought/apptools/pull/261
If the name of "build" sounds good to you, the file can be added here regardless of whether that PR or this PR is merged first. If #261 is not backported to the maintenance branch, then we will see the leftover news fragments being detected when the changelog is built again prior to the final release... but it is probably most convenient to backport the change in #261.

README.rst
```diff
@@ -37,10 +37,39 @@ All packages in apptools require:
...
 
+Certain sub-packages within apptools have their own specific dependencies,
+which are optional for apptools overall.
+
 The `apptools.preferences` package requires:
 
 * `configobj <http://pypi.python.org/pypi/configobj>`_
 
+To install with `apptools.preferences` dependencies::
+
```
kitchoi, line 48: Thank you for adding these. 

Do you mind pulling this out to a separate installation section? It interleaving with the list of dependencies makes this section difficult to parse at a glance.

Nit: The single backtick does not change the rendering styling. To match the styling in the first section, using bold text, e.g. `**apptools.preferences**`, might be a neater.

aaronayres35: I ended up leaving the install instructions as code blocks matching the pyface example.  I think with install instructions in a separate section this looks clean, but I am happy to change to the ** ** approach if you prefer it

kitchoi (commented):
LGTM. Thank you.
Shall we have a news fragment for this change? https://github.com/enthought/apptools/pull/261 is motivated by this PR, but I thought I should confirm with you first before merging it.

aaronayres35:
> LGTM. Thank you.
> Shall we have a news fragment for this change? #261 is motivated by this PR, but I thought I should confirm with you first before merging it.

I've added one in the latest commit, I just used the `build` name before 261 was merged

kitchoi (approved):
> I've added one in the latest commit, I just used the build name before 261 was merged.


Oops. I am sorry. I failed refresh the tab 🤦

docs/releases/upcoming/257.build.rst
```diff
@@ -0,0 +1 @@
+Add extras_require to setup.py for optional dependencies
```
kitchoi, line 1: You will need `(#257)` at the end.

PR #260, opened 2020-12-10
Make sure optional dependencies are optional for tests

### LINKED ISSUE #99: Make sure that optional dependencies are also optional for tests

aaronayres35:
fixes #99 

This PR uses the existing `apptools._testing.optional_dependencies` to ensure that optional dependencies are optional for tests.  The test suite should no longer fail if certain optional dependencies are not installed.

**Checklist**

### DIFF
--- a/apptools/persistence/tests/test_state_pickler.py
+++ b/apptools/persistence/tests/test_state_pickler.py
@@ -19,7 +19,10 @@ import math
 import os
 import tempfile
 
-import numpy
+from apptools._testing.optional_dependencies import (
+    numpy,
+    requires_numpy,
+)
 
 from traits.api import (
     Bool,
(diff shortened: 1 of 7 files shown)


apptools/persistence/tests/test_state_pickler.py
```diff
@@ -43,69 +46,68 @@
...
-        self.dict = {"a": 1, "b": 2, "ref": self.inst}
-        self.ref = self.numeric
+            _tvtk = Instance(tvtk.Property, ())
+
+        def __init__(self):
+            self.inst = A()
+            self.tuple = (1, 2, "a", A())
+            self.list = [1, 1.1, "a", 1j, self.inst]
+            self.dict = {"a": 1, "b": 2, "ref": self.inst}
+            self.ref = self.numeric
```
aaronayres35, line 107: This was a bit awkward, but these classes should only be defined if we have numpy, so I moved them inside the above if statement.  There is probably a cleaner way to do this

aaronayres35:
I am curious to test the new news fragment system with this scenario, as this is a PR that will need to be backported for the 5.0.0 release and it has a news fragment.

apptools/persistence/tests/test_state_pickler.py
kitchoi: Since all of the tests in this module requires NumPy, this is a case where raising `SkipTest` at the module is appropriate, like this: https://github.com/enthought/traitsui/blob/464b20200eef30f23317ffd3f6a125520329f7e2/traitsui/tests/ui_editors/test_data_frame_editor.py#L11-L14

With that, when we run the test with discover, the entire module will be skipped.  And if one wants to run the test module on its own, e.g. `python -m unittest apptools.persistence.tests.test_state_pickler`, they ought to have NumPy because the subject `state_pickler` requires NumPy. It is understandable to get an exception there.

kitchoi: I think I flip flop between the two (wrapping code in a condition VS raising SkipTest at the module) depending on the situation (e.g. how optional the dependency is, how much code needs to be wrapped in an if-block, ...). This is an option for consideration.

kitchoi: The approach here means in both run scenarios (`unittest discover`) and (`unittest <module-name>`), the tests are consistently skipped in the absence of numpy.

kitchoi (approved):
LVGTM. Thank you.

There is an alternative way to skip the entire test module for you consider. I don't have a particular preference over which one.

apptools/persistence/tests/test_state_pickler.py
aaronayres35: In this particular case I went with the approach you described as it is cleaner IMO.  For the rest I left them as is.

aaronayres35:
@kitchoi still looking okay to merge?

kitchoi:
> still looking okay to merge?

Yes. Thank you!

PR #263, opened 2020-12-16
Backports for 5.0.0 and update Changelog

aaronayres35:
[Targeting maint/5.0]

This PR backports merged PRs to the 5.0 branch in preparation for the upcoming release. It also updates the changelog to include these backported PRs. Only the last commit which updates the changelog needs to be reviewed (the others have been reviewed already).

**Checklist**


CHANGES.txt
```diff
@@ -52,6 +52,11 @@ Documentation changes
 Test suite
 ----------
 * Fix AttributeError on Python 3.9 due to usage of ``base64.decodestring`` in tests (#210)
+* Make optional dependencies optional for tests (#260)
```
aaronayres35, line 55: One thing I noticed in this PR (it very well may have already been pointed out, I can't remember), is that with the new news fragment system PRs are added to the changelog in order of lowest to highest PR number, which may differ from other ets changelogs.
Not saying anything needs to change, just an observation

rahulporuri (approved):
LGTM

CHANGES.txt
rahulporuri: Ahh. Good point. Not anything we need to change though.

PR #269, opened 2020-12-31
Use templates to build api docs

rahulporuri:
This PR uses templates to build API documentation. The work has been lifted from traits-futures https://github.com/enthought/traits-futures/pull/181/

|before|after|
|-------|-----|
|apptools-before|apptools-after|

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~


docs/source/api/apptools.api.rst
```diff
@@ -0,0 +1,15 @@
+..
+   (C) Copyright 2006-2020 Enthought, Inc., Austin, TX
```
kitchoi, line 2: 2️⃣ 0️⃣ 2️⃣ 1️⃣  🎉

docs/source/index.rst
```diff
@@ -11,6 +11,14 @@ AppTools Documentation
...
     naming/*
     io/*
-    api
+
+API Documentation
+-----------------
+
+.. toctree::
+    :maxdepth: 1
+    :glob:
```
kitchoi, line 20: Is this `:glob:` needed?
(I believe the `:glob:` in the above section is for the use of `*`. But here `*` is not used.)

kitchoi (commented):
I saw the following warnings when I build the documentation after doing a `git clean`
```
/Users/kchoi/Work/ETS/apptools/docs/source/api.rst:8: WARNING: toctree contains reference to nonexisting document 'api/modules'
WARNING: autodoc: failed to import module 'api' from module 'apptools'; the following exception was raised:
No module named 'apptools.api'
```
When I build the documentation from master, I don't see the same warning.

Looks like that file can be removed (it might break the URL a bit for that particular page, probably not a big deal.)

kitchoi (approved):
LGTM.
(Haven't manually built the documentation with the latest changes.)

PR #272, opened 2021-01-08
Use pyface.undo imports in apptools.undo instead of redefining the classes

### LINKED ISSUE #271: Use pyface.undo in apptools.undo
At the moment, `apptools.undo` defines all of the relevant classes but raises a deprecating warning - and recommends users to use `pyface.undo` instead of the relevant `apptools.undo` modules.

While this is a good place to start, this is not enough. What we really want is to remove all definitions of classes from `apptools.undo` modules and simply import them from `pyface.undo`. This will ensure that the code remains backwards compatible if projects use `apptools.undo.*` objects together with `pyface.undo.*` objects, making it easy to update their imports.

rahulporuri:
fixes #271 

This PR imports the relevant classes from `pyface.undo` and `pyface.undo.action` instead of redefining the classes. Where possible, we import from `pyface.undo.api` and `pyface.undo.action.api`.

Note that the tests have been removed as this module is now simply a stub, exposing the functionality in `pyface.undo`, and will be removed in the future.

**Checklist**


corranwebster (approved):
Looks good to me.

apptools/undo/abstract_command.py
```diff
@@ -12,65 +12,4 @@
...
-        from the original call to 'do()' or previous call to 'redo()'.
-        """
-
-        raise NotImplementedError
-
-    def undo(self):
-        """ This is called by the command stack to undo the command. """
-
-        raise NotImplementedError
+from pyface.undo.api import AbstractCommand
```
kitchoi, line 15: For maintaining backward compatibility, I'd think the safest thing to do would be `from pyface.undo.abstract_command import *`, assuming the module is moved to pyface as is. e.g. this breaks `from apptools.undo.abstract_command import ICommand`. We believe no one does that, but it costs little to do the safer approach. 🤞  this does not break anything.

aaronayres35: @rahulporuri should we make this change before I push ahead with the release?

rahulporuri: i'd actually prefer breaking such imports now - which will make the eventual transition to `pyface` trivial - and cleaner/better because those imports are antipatterns anyway.

PR #273, opened 2021-01-11
Correct RedoAction/UndoAction imports from pyface

rahulporuri:
This fixes an issue introduced in the PR #272 - `apptools.undo.action.api` tries importing `RedoAction`/`UndoAction` from `pyface.undo.action.command_action` - which is not the right thing to do. This import has been updated to use `pyface.undo.action.api` instead.

This also fixes an new sphinx warning introduced in #272 

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~

### DIFF
--- a/apptools/undo/action/api.py
+++ b/apptools/undo/action/api.py
@@ -12,7 +12,7 @@
 # Description: <Enthought undo package component>
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

PR #274, opened 2021-01-11
Add docstrings to api modules in apptools subpackages

rahulporuri:
This PR adds module docstrings to all api modules in the various apptools subpackages.

**Checklist**


aaronayres35 (approved):
Changes LGTM

One thing I noticed though (This is probably already known), is that when you navigate to the page for one of the api modules rather than just seeing links to the docs for the objects you get the api docs directly on that page.  In other words, the full contents for the docs of the `apptools.naming` package are on one page.  Consider this example from envisage:
Screen Shot 2021-01-11 at 8 07 58 AM

Whereas in apptools we have:
Screen Shot 2021-01-11 at 8 09 27 AM
Notice the section for `apptools.naming.binding module` at the bottom of the screenshot.  There are sections for the other naming submodules as well all on this same page.

This may be intentional / unrelated to the changes of this PR so feel free to ignore, but I wanted to make sure it was pointed out

rahulporuri:
> This may be intentional / unrelated to the changes of this PR so feel free to ignore, but I wanted to make sure it was pointed out

Yup. This is because of the recent change where we use templates when building docs.

PR #285, opened 2021-03-30
Replace on_trait_change with observe

aaronayres35:
**Checklist**

This PR replaces _almost_ all uses of `on_trait_change` with observe.

All that remains is `apptools.selection.selection_service.py` and `apptools.scripting.recorder.py` (also an undo example but that is deprecated here anyway).  These are a bit trickier and can be addressed separately

### DIFF
--- a/apptools/preferences/preference_binding.py
+++ b/apptools/preferences/preference_binding.py
@@ -69,10 +69,10 @@ class PreferenceBinding(HasTraits):
 
     #### Trait change handlers ################################################
 
-    def _on_trait_changed(self, obj, trait_name, old, new):
+    def _on_trait_changed(self, event):
         """ Dynamic trait change handler. """
 
-        self.preferences.set(self.preference_path, new)
+        self.preferences.set(self.preference_path, event.new)
 
     #### Other observer pattern listeners #####################################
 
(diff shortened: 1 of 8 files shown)


apptools/preferences/ui/widget_editor.py
```diff
@@ -58,14 +58,14 @@ def update_editor(self):
     # Private interface.
     ###########################################################################
 
-    def _on_page_changed(self, obj, trait_name, old, new):
+    def _on_page_changed(self, event):
```
rahulporuri, line 61: Maybe rename to `_handle_change_to_selected_page`?

Also, i'm not sure why we're not using the decorator here,

rahulporuri: It's quite possible that this code was written before the decorator existed in traits - i don't see a good reason why we can't use the new decorator here.

rahulporuri (approved):
LGTM with a few comments/suggestions.

apptools/preferences/preference_binding.py
```diff
@@ -121,7 +121,7 @@ def _initialize(self):
         """ Wire-up trait change handlers etc. """
 
         # Listen for the object's trait being changed.
-        self.obj.on_trait_change(self._on_trait_changed, self.trait_name)
+        self.obj.observe(self._on_trait_changed, self.trait_name)
```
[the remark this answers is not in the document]
aaronayres35: yeah I noticed this as well... not sure what the best thing to do is

aaronayres35:
Orthogonal to this PR, but we should add a "maintenance" news fragment type. I wasn't sure which of "feature, bugfix, deprecation, removal, doc, test, build" to classify this PR as, but it feels like it will be ultimately relevant for the changelog (as I imagine 5.1 of apptools main purpose will be this migration to observe / depending on traits 6.2)

I am going to call it a feature for now simply because IDK where to put it. When generating the actual changelog we will want to move this, but this way it will still be picked up

PR #286, opened 2021-04-16
Replace use of envisage "contributes_to" decorator in docs

rahulporuri:
This PR removes the use of envisage `@contributes_to` decorator in favor of a trait `List` with metadata `contributes_to` instead. This is being done because the next release of envisage removes the decorator entirely. Ref https://github.com/enthought/envisage/pull/402

### DIFF
--- a/docs/source/preferences/PreferencesInEnvisage.rst
+++ b/docs/source/preferences/PreferencesInEnvisage.rst
@@ -38,11 +38,16 @@ loaded into the default scope when the application is started.
 
 e.g. To contribute a preference file for my plugin I might use::
 
+  from envisage.ids import PREFERENCES
+  ...
+
+
   class MyPlugin(Plugin):
       ...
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

This files existence entirely is a bit out of place.  There are preferences docs in the envisage docs that we likely want to merge with this (https://docs.enthought.com/envisage/envisage_core_documentation/preferences.html)

rahulporuri:
> Orthogonal to this PR, but this files existence entirely is a bit out of place. There are preferences docs in the envisage docs that we likely want to merge with this (https://docs.enthought.com/envisage/envisage_core_documentation/preferences.html)

You're right. These docs do probably belong in envsiage, not in apptools.

rahulporuri:
created https://github.com/enthought/envisage/issues/403

Issue #287, opened 2021-06-14
Use GitHub Actions CI instead of Travis CI


rahulporuri:
Along with the rest of ETS, we would like to move apptools to use GitHub Actions CI instead of Travis CI.

PR #288, opened 2021-06-14
Port CI to gh-actions

aaronayres35:
Part of #287 (which can be closed when we drop travis and appveyor)
This PR adds two _nearly_ identical work flows.  One of which will run the test suite and flake8 on every PR using packages from edm, the other which will run the test suite as a cron job using ets packages from source. 

**Checklist**


rahulporuri (approved):
LGTM

rahulporuri:
I dont think this PR needs a changelog entry

PR #289, opened 2021-06-14
Replace Property depends_on with observe

aaronayres35:
Note these changes now make `apptools` require traits >= 6.2.  ~I checked to specify this requirement, but it appears as though traits isn't even explicitly listed as a requirement, but rather is implicitly required through `traitsui` (which now explicitly requires traits >= 6.2).~ EDIT: I've added `traits>=6.2.0` as an explicit required dependency on this PR.

https://github.com/enthought/apptools/blob/96c3af4213f0d3ab2c831f4e2949a5170d6ff96c/setup.py#L305-L308

**Checklist**

### DIFF
--- a/apptools/logger/plugin/view/logger_view.py
+++ b/apptools/logger/plugin/view/logger_view.py
@@ -95,10 +95,10 @@ class LoggerView(TraitsUIView):
     service = Instance(LoggerService)
 
     log_records = List(Instance(logging.LogRecord))
-    formatted_records = Property(Str, depends_on="log_records")
+    formatted_records = Property(Str, observe="log_records")
 
     activated = Instance(logging.LogRecord)
-    activated_text = Property(Str, depends_on="activated")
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM

aaronayres35:
Agh, I forgot a news fragment. I will open a separate PR.

rahulporuri:
@aaronayres35 dunno if this is worth a news fragment - well maybe a news fragment mentioning that we now explicitly depend on traits >= 6.2.0.

aaronayres35:
> @aaronayres35 dunno if this is worth a news fragment - well maybe a news fragment mentioning that we now explicitly depend on traits >= 6.2.0.

I guess that is true, we have had changelog entries for otc-> observe changes in other packages, but they typically contain a list of all PRs involved, eg depends_on, dynamic and static handlers, _name_changed, etc.  I guess I will just close #952.

PR #299, opened 2021-09-17
Replace `eval` with `ast.literal_eval` in apptools.preferences

corranwebster:
We don't need the full power of `eval()` here, `literal_eval()` should do everything we need, is safer, and anything which breaks can be modified to use a custom trait that is more focused.  This change doesn't break any existing tests, but includes a new test that something which should fail due to the change, does.

**Checklist**


jwiggins (approved):
LGTM.

PR #306, opened 2022-05-14
Use context managers with preferences lock.

corranwebster:
A drive-by code modernization to replace acquire()/release() pairs with the safer use of a context manager.

Includes a drive-by modernization of the super() statement.

Should be no change to behaviour _except_ if something goes wrong, and then it will now do the right thing and release the lock.

**Checklist**


corranwebster:
Needs #307 for flake8 to pass.

mdickinson (approved):
LGTM

Issue #303, opened 2022-01-13
Test failures on Python 3.10


fabaff:
With Python 3.10 some tests are failing.

```bash
============================= test session starts ==============================
platform linux -- Python 3.10.1, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /build/apptools-5.1.0
collected 277 items / 3 errors / 274 selected                                  

==================================== ERRORS ====================================
___________ ERROR collecting apptools/io/h5/tests/test_dict_node.py ____________
ImportError while importing test module '/build/apptools-5.1.0/apptools/io/h5/tests/test_dict_node.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/nix/store/b798fp24zf2fdafmyyc4sxfr48ly5yy9-python3-3.10.1/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
apptools/io/h5/tests/test_dict_node.py:21: in <module>
    from .utils import open_h5file, temp_h5_file, temp_file
apptools/io/h5/tests/utils.py:14: in <module>
    from ..utils import open_h5file
apptools/io/h5/utils.py:12: in <module>
    from .file import H5File
apptools/io/h5/file.py:10: in <module>
    from collections import Mapping, MutableMapping
E   ImportError: cannot import name 'Mapping' from 'collections' (/nix/store/b798fp24zf2fdafmyyc4sxfr48ly5yy9-python3-3.10.1/lib/python3.10/collections/__init__.py)
______________ ERROR collecting apptools/io/h5/tests/test_file.py ______________
ImportError while importing test module '/build/apptools-5.1.0/apptools/io/h5/tests/test_file.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/nix/store/b798fp24zf2fdafmyyc4sxfr48ly5yy9-python3-3.10.1/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
apptools/io/h5/tests/test_file.py:22: in <module>
    from ..file import H5File
apptools/io/h5/file.py:10: in <module>
    from collections import Mapping, MutableMapping
E   ImportError: cannot import name 'Mapping' from 'collections' (/nix/store/b798fp24zf2fdafmyyc4sxfr48ly5yy9-python3-3.10.1/lib/python3.10/collections/__init__.py)
[...]

PR #314, opened 2022-07-28
Add workflow that tests against pip packages for various Python versions

mdickinson:
This PR adds a simple GitHub actions workflow, triggered on pull request (and optionally also manually) that runs the test suite on Python 3.6, 3.8 and 3.10 in a standard Python environment (rather than an EDM environment). The goal is to catch issues like #303 earlier.

---

EDIT: PR updated to also include a `pyproject.toml` file.


rahulporuri (approved):
LGTM

PR #315, opened 2022-08-15
Add workflow to publish releases to PyPI

mdickinson:
This PR adds a workflow which automatically publishes releases to PyPI.

This is almost identical to the corresponding workflow on Envisage (which ran successfully this morning). The only difference is an extra line to upgrade `pip` if necessary. (This line doesn't make a difference right now, but it's not inconceivable that some future release of `build` or `twine` can only be installed properly with an updated `pip`.)

I've tested the commands in the workflow manually by:

- Creating a new, clean Python 3.10 venv
- Running `python -m pip install --upgrade pip` and `python -m pip install build twine`
- Running `python -m build` followed by `python -m twine check --strict dist/*`.

The strict checks passed for me.


mdickinson:
Note: appropriate `PYPI_USERNAME` and `PYPI_PASSWORD` secrets have been added to the repository.

docs/releases/upcoming/313.build.rst
```diff
@@ -0,0 +1 @@
+Add GitHub Actions workflow for automatically publishing releases to PyPI.
```
mdickinson, line 1:
```suggestion
Add GitHub Actions workflow for automatically publishing releases to PyPI. (#315)
```

corranwebster (approved):

PR #316, opened 2022-08-15
Update etstool for use with Python 3.8

mdickinson:
This PR updates `etstool.py` and the EDM-based test workflow to support Python 3.8 as well as Python 3.6.

- Updates etstool.py
- Updates the EDM-based PR-test workflow
- Updates the EDM-based test-with-ETS-source cron-job workflow

The `etstool.py` changes ended up being more extensive than I originally intended, since I ran into issues with using a non-EDM Python as a bootstrap environment on Windows. So there's now some additional logic to identify the EDM executable on Windows, overridable through a command-line argument or an environment variable. (This logic was copied from other similar workflows, like the one in Envisage.)

**Checklist**


mdickinson:
(successful) ets-from-source run on this branch: https://github.com/enthought/apptools/actions/runs/2860939442

corranwebster (approved):

Issue #323, opened 2022-10-17
LogFileHandler does not support encoding


mechworrior:
Currently `LogFileHandler` does not inherit the `encoding` attribute from `RotatingFileHandler`. 
This is required to enable `utf-8` characters in Logs.

rahulporuri:
It will be good to confirm that all of the kwargs accepted by `RotatingFileHandler` are also accepted by the `LogFileHandler` and that they are passed through as necessary.

mdickinson:
This was fixed in #324.

PR #324, opened 2022-10-17
FIX: add encoding as an attribute for LogFileHandler initialization

mechworrior:
Fix for #323 

Simply adds `encoding` as attribute for `LogFileHandler`.

### DIFF
--- a/apptools/logger/logger.py
+++ b/apptools/logger/logger.py
@@ -29,7 +29,13 @@ class LogFileHandler(RotatingFileHandler):
     """The default log file handler."""
 
     def __init__(
-        self, path, maxBytes=1000000, backupCount=3, level=None, formatter=None, encoding=None
+        self,
+        path,
+        maxBytes=1000000,
+        backupCount=3,
+        level=None,
+        formatter=None,
+        encoding=None,
     ):
         RotatingFileHandler.__init__(
             self, path, maxBytes=maxBytes, backupCount=3, encoding=encoding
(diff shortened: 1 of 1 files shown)


apptools/logger/logger.py
```diff
@@ -29,10 +29,10 @@ class LogFileHandler(RotatingFileHandler):
     """The default log file handler."""
 
     def __init__(
-        self, path, maxBytes=1000000, backupCount=3, level=None, formatter=None
+        self, path, maxBytes=1000000, backupCount=3, level=None, formatter=None, encoding=None
     ):
         RotatingFileHandler.__init__(
-            self, path, maxBytes=maxBytes, backupCount=3
+            self, path, maxBytes=maxBytes, backupCount=3, encoding=encoding
```
rahulporuri, line 35: Let's fully support the existing API - `RotatingFileHandler(filename, mode='a', maxBytes=0, backupCount=0, encoding=None, delay=False)`

apptools/logger/logger.py
```diff
@@ -29,10 +29,24 @@ class LogFileHandler(RotatingFileHandler):
...
+        backupCount=3,
+        level=None,
+        formatter=None,
+        encoding=None,
+        delay=False,
     ):
         RotatingFileHandler.__init__(
-            self, path, maxBytes=maxBytes, backupCount=3
+            self,
+            path=path,
```
rahulporuri, line 44:
```suggestion
            filename=path,
```

It looks like the kwarg is called `filename` on the `RotatingFileHandler`, not `path` like what we're using - https://github.com/python/cpython/blob/8d999cbf4adea053be6dbb612b9844635c4dfb8e/Lib/logging/handlers.py#L122

rahulporuri (approved):
🎉  LGTM

PR #331, opened 2023-03-23
Port `Workbench` from Pyface to Apptools

corranwebster:
This is a move of the workbench code from Pyface to Apptools.  The top-level changes are:

- `pyface.workbench` -> `apptools.workbench`
- `pyface.ui.*.workbench` -> `apptools.workbench.ui.*` (I could be talked into just `apptools.workbench.*` like TraitsUI)
- `examples.workbench` in Pyface repo added to `examples.workbench` in this repo
- adds some tests run under PySide6 and Wx

Toolkit management is done by using a `pyface.base_toolkit.Toolkit` with the name derived from the `pyface.toolkit.toolkit` (which handles toolkit selection) and pointing at the appropriate subpackage.

This doesn't attempt to do anything to the Workbench code in Envisage.  There is an argument it should be in here as well.  We have circular dependencies between the two libraries, and it's not clear which should be "lower-level"

Other than that, changes are around import targets and a little bit of delayed importing because I was seeing circular imports after applying isort...

**Checklist**
- [x] Check examples work
- [x] Run tests under different backends

### DIFF
--- a/docs/releases/upcoming/331.feature.rst
+++ b/docs/releases/upcoming/331.feature.rst
@@ -1,0 +1,1 @@
+Move Workbench to Apptools, as advertised in Pyface (#331)


corranwebster:
Currently not working with an actual toolkit.

PR #338, opened 2024-06-27
Use a temporary file for the application_preferences_filename

### LINKED ISSUE #337: Test suite fails if .enthought/unittest/preferences.ini already exists
Steps to reproduce (Linux / macOS):

- create a `.enthought/unittest/preferences.ini` file with the following contents:

```console
$ cat ~/.enthought/unittest/preferences.ini
[app]
server_url = https://my.server.com
```

- run the test suite

Then (for me), the test suite fails with the following error:

```
FAIL: test_node_names (apptools.preferences.tests.test_scoped_preferences.ScopedPreferencesTestCase.test_node_names)
node names
  File "/Users/mdickinson/Enthought/ETS/apptools/apptools/preferences/tests/test_preferences.py", line 242, in test_node_names
    self.assertEqual([], p.node_names())
AssertionError: Lists differ: [] != ['app']

Second list contains 1 additional elements.
First extra element 0:
'app'

- []
+ ['app']
```

mdickinson:
Closes #337 

This PR ensures that tests of `ScopedPreferences` use a temporary file for the application preferences scope.


apptools/preferences/tests/test_scoped_preferences.py
```diff
@@ -42,14 +42,18 @@ class ScopedPreferencesTestCase(PreferencesTestCase):
     def setUp(self):
         """ Prepares the test fixture before each test method is called. """
 
-        self.preferences = ScopedPreferences()
+        # A temporary directory that can safely be written to.
+        self.tmpdir = tempfile.mkdtemp()
```
flongford, line 46: _Side comment_: should we consider creating an issue to replace all `tempfile.mkdtemp` and `tempfile.mktemp` with more modern `tempfile.TemporaryFile` and `tempfile.TemporaryDirectory` routines?

flongford (approved):
LGTM!

apptools/preferences/tests/test_scoped_preferences.py
mdickinson: I'm not seeing a reason to change uses of `tempfile.mkdtemp`; for sure it would be nice to have, but there's nothing wrong with `mkdtemp` and general code modernization isn't really something we have bandwidth for. Uses of `tempfile.mktemp` would definitely need to be fixed, since it's deprecated and insecure, but I'm not seeing any uses in the codebase.

PR #339, opened 2024-06-27
Drop Python 3.7 support and fix test workflows

mdickinson:
This PR brings our CI jobs back to working state, and extends them to cover more recent Python versions:

- Support for Python 3.7 is dropped.
- `etstool.py` has been updated to support Python 3.11
- the `test-with-pip.yml` workflow has dropped 3.7 and added 3.12
- the `test-with-edm.yml` workflow has dropped 3.6 and added 3.11
- a `dependabot.yml` config has been added, just for GitHub Actions
- limit NumPy to versions < 2.0 for now - PyTables currently doesn't like NumPy 2.0
- don't install PyTables on Apple Silicon - it's not currently supported there

**Checklist**


etstool.py
```diff
@@ -101,19 +101,27 @@ def edm_dependencies(runtime):
...
+        "tables",
+        "traitsui",
     }
+    runtime_specific_dependencies = {
+        "3.8": {
+            # Most of these are currently unavailable on Python 3.11;
+            "enthought_sphinx_theme",
+            "flake8",
+            "flake8_ets",
+            "sphinx",
```
flongford, line 118: Again, just for context, I see that we don't have any CI jobs that build documentation, so are we using Sphinx anywhere?

flongford (approved):
LGTM! Just some minor questions for additional context

etstool.py
mdickinson: We use Sphinx to build docs after making a release, and the current workflow for that is to do it manually via etstool.

mdickinson: Agreed that we _should_ ideally have a workflow for that.

.github/workflows/test-with-pip.yml
```diff
@@ -20,9 +20,13 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies and local packages
-      run: |
-        python -m pip install .[h5,preferences]
+    - name: Install dependencies and local packages (not macOS)
+      run: python -m pip install .[h5,preferences]
+      if: matrix.os != 'macos-latest'
+    - name: Install dependencies and local packages (macOS)
+      run: python -m pip install .[preferences]
+      # tables currently won't build on Apple Silicon
```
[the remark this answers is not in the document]
mdickinson: I'm not sure whether it's really worth an issue here - I don't see that there's anything actionable for apptools other than wait until PyTables figures out how to support Apple Silicon. The apptools code should all work the moment that that happens.

The most relevant upstream issue is here: https://github.com/PyTables/PyTables/issues/1165

I should probably at least link to that in the workflow.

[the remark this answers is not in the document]
mdickinson: > Ah, it's the latter I see

Right: PyTables is in effect an optional dependency. Many of our uses of apptools only care about some other part (e.g., preferences) and don't care about the persistence pieces.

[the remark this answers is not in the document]
mdickinson: > I'm not sure whether it's really worth an issue here

Okay, I take that back - there is an action to be taken, which is to remove the workaround from our workflows. I'll open issues both for this and for the NumPy 2.0 incompatibility.

etstool.py
flongford: Got it, thanks!

PR #341, opened 2024-06-27
Remove uses of numpy.alltrue

mdickinson:
This PR removes uses of `numpy.alltrue`, which no longer exists in NumPy 2.0.

**Checklist**


mdickinson:
Self-merging (with some feelings of guilt).

PR #343, opened 2024-06-27
Avoid writing to preferences on helper creation

### LINKED ISSUE #74: Preferences helper will call `_initialize` twice at construction
The current implementation will call `_initialize` during the HasTraits contructor and at the same time the static change handler on the `preferences` trait would have already triggered the same call when the parent constructor is called.

While this behavior is normally harmless it is problematic and unnecessary.  

https://github.com/enthought/apptools/blob/94c4963f8aa9ea5d52f1d5c8020df54db1c96524/apptools/preferences/preferences_helper.py#L39

### LINKED ISSUE #342: Preferences written on helper creation
Preferences are currently written the first time a `PreferencesHelper` for those preferences is created. This leads to some surprising behaviour, especially in scoped preferences. Here's an example test case that currently fails:

```python
    def test_preferences_not_written_on_helper_creation(self):

        class AppPreferencesHelper(PreferencesHelper):
            #: The node that contains the preferences.
            preferences_path = "app"

            #: The user's favourite colour
            color = Str()

        default_preferences = Preferences(name="default")
        default_preferences.set("app.color", "red")

        application_preferences = Preferences(name="application")
        preferences = ScopedPreferences(
            scopes=[application_preferences, default_preferences]
        )
        self.assertIsNone(application_preferences.get("app.color"))

        # Then creation of the helper should not cause the application
        # preferences to change.
        AppPreferencesHelper(preferences=preferences)
        self.assertIsNone(application_preferences.get("app.color"))
```

This test case fails at the second `self.assertIsNotNone` line: the preferences from the default scope have been written into the application scope.

mdickinson:
This PR fixes the `PreferencesHelper` to avoid writing to the preferences the first time a helper is created for a given `Preferences` object.

The issue was that on creation, `PreferencesHelper._initialize` was being called twice: once with `notify=False` and again with `notify=True`, as a result of the listener to the `preferences` trait. This PR avoids the second call by making that listener only run post creation (using `post_init=True`).

Closes #342
Closes #74

**Checklist**

### DIFF
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -100,7 +100,7 @@ class PreferencesHelper(HasTraits):
         # the trait is not a preference trait and we do nothing.
 
     @observe("preferences", post_init=True)
-    def _preferences_changed(self, event):
+    def _update_preferences_listeners(self, event):
         """Update listeners when the preferences object changes."""
 
         old, new = event.old, event.new
--- a/apptools/preferences/preferences_helper.py
+++ b/apptools/preferences/preferences_helper.py
@@ -15,7 +15,7 @@ from ast import literal_eval
 import logging
 
 # Enthought library imports.
-from traits.api import HasTraits, Instance, Str
+from traits.api import HasTraits, Instance, Str, observe
 
 # Local imports.
 from .i_preferences import IPreferences
@@ -99,8 +99,11 @@ class PreferencesHelper(HasTraits):
         # If the change refers to a trait defined on this class, then
         # the trait is not a preference trait and we do nothing.
 
-    def _preferences_changed(self, old, new):
-        """ Static trait change handler. """
+    @observe("preferences", post_init=True)
+    def _preferences_changed(self, event):
+        """Update listeners when the preferences object changes."""
+
+        old, new = event.old, event.new
 
         # Stop listening to the old preferences node.
         if old is not None:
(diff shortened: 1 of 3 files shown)


mdickinson:
@flongford Would you be able to review?

flongford (approved):
LGTM. The use of an additional `observe` listener seems appropriate here and appears to resolve the reported issue.

I can't think of any immediate downstream consequences of this change, but I expect these can be simply handled in a case-by-case basis.

apptools/preferences/preferences_helper.py
```diff
@@ -99,8 +99,11 @@ def _anytrait_changed(self, trait_name, old, new):
         # If the change refers to a trait defined on this class, then
         # the trait is not a preference trait and we do nothing.
 
-    def _preferences_changed(self, old, new):
-        """ Static trait change handler. """
+    @observe("preferences", post_init=True)
+    def _preferences_changed(self, event):
```
mdickinson, line 103: Hmm. To avoid confusion, the name of this method should be changed away from the magic method name.
```suggestion
    def _update_preferences_listeners(self, event):
```

mdickinson: Fixed in [6cb93e8](https://github.com/enthought/apptools/pull/343/commits/6cb93e8ba3d086650be9bee13a111205df5d0a34)

PR #350, opened 2024-06-28
Remove slack notification action

mdickinson:
The Slack notification action that we were using is no longer supported, and there isn't an obvious good replacement. This PR removes it.


mdickinson:
Trivial PR that doesn't touch apptools business logic; self-merging.

PR #86, opened 2019-02-03
Treating traitsui as optional

stewbond:
If users don't have traitsui, then GUI-specific functions now do nothing instead of crashing the application.  This really helps for running unit tests on build machines without any X-server which have problems when running traitsui functions.

Note that there probably some other places where traitsui could be removed, but this at least gets the CI tests working.

Also helps with issue #70.

PR #351, opened 2024-06-28
Make TraitsUI, Pyface and configobj optional dependencies

### LINKED ISSUE #70: Consider making traitsui an optional requirement
Projects that don't use the GUI parts of apptools are still required to install traitsui if they want to have an environment where all the dependencies are met:

```
09:56 $ edm install apptools
Fetching indices for package repositories.... done

The following packages will be installed:

    apptools   4.4.0-4   787     KiB
    configobj  5.0.6-2    70     KiB
    pyface     5.1.0-5     1.5   MiB
    pygments   2.1.3-1     1.73  MiB
    six        1.10.0-1   19     KiB
    traits     4.6.0-1   636     KiB
    traitsui   5.1.0-2     1.5   MiB

Do you want to continue ? [Y/n]:
```

mdickinson:
Apptools contains several disparate packages with different needs. While it's somewhat reasonable to regard Traits as a core dependency, not all subpackages need a GUI, and only the preferences package makes use of configobj.

This PR:

* makes all three of TraitsUI, Pyface and configobj optional dependencies
* adds a "gui" entry for `extras_require` in `setup.py`
* adds a test workflow that runs the test suite with only the core dependencies (i.e., Traits), to make sure that other tests are skipped appropriately rather than failing

Closes #70 

Makes #86 obsolete.

**Checklist**

### DIFF
--- a/apptools/logger/plugin/tests/test_logger_service.py
+++ b/apptools/logger/plugin/tests/test_logger_service.py
@@ -11,9 +11,13 @@ from email.mime.multipart import MIMEMultipart
 import unittest
 from unittest import mock
 
-from apptools.logger.plugin.logger_service import LoggerService
+from apptools._testing.optional_dependencies import pyface, requires_pyface
+
+if pyface is not None:
+    from apptools.logger.plugin.logger_service import LoggerService
 
 
+@requires_pyface
 class LoggerServiceTestCase(unittest.TestCase):
     def test_create_email_message(self):
         logger_service = LoggerService()
--- a/setup.py
+++ b/setup.py
@@ -302,15 +302,15 @@ if __name__ == "__main__":
                 'preferences/tests/*.ini'
             ]
         },
-        install_requires=[
-            'configobj',
-            'traits>=6.2.0',
-            'traitsui',
-        ],
+        install_requires=['traits>=6.2.0'],
         extras_require={
             "docs": ["enthought-sphinx-theme", "sphinx"],
             "test": [
                 "importlib-resources>=1.1.0; python_version<'3.9'",
+            ],
+            "gui": [
+                "pyface",
+                "traitsui",
             ],
             "h5": [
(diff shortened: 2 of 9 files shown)


apptools/logger/plugin/tests/test_logger_service.py
```diff
@@ -11,9 +11,13 @@
...
 from unittest import mock
 
-from apptools.logger.plugin.logger_service import LoggerService
+from apptools._testing.optional_dependencies import pyface, requires_pyface
 
+if pyface is not None:
+    from apptools.logger.plugin.logger_service import LoggerService
 
+
+@requires_pyface
```
flongford, line 20: _Side comment:_ I'm kind of surprised that this is the only place we're testing `pyface` functionality, but I guess that's because we're ignoring all the `apptools.undo` module unit tests.

mdickinson:
> I presume that these updated dependencies will be propagated through to our "publish on pip" GH actions workflow without further changes required.

Yes - the `setup.py` should take care of that.

> Do we have any "publish on EDS" equivalent workflows (automated or manual) that need to be updated post-merge?

We don't: the "publish on EDS" workflow is entirely manual (I go and fill out a build request). So that build request will include a note that the requirements have changed.

apptools/logger/plugin/tests/test_logger_service.py
mdickinson: > because we're ignoring all the `apptools.undo` module unit tests.

Wait - we have tests for apptools.undo?!

flongford: Sorry, my mistake - I saw the `apptools.undo` module still exists and didn't realise that it wasn't populated with unit tests.

In which case, this makes complete sense

setup.py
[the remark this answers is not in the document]
flongford: Makes sense to me, thanks for the explanation

flongford (approved):

PR #353, opened 2024-07-04
Prepare for release of version 5.3.0 of apptools

mdickinson:
This PR (against the `main` branch) bumps the version and updates the changelog in preparation for a 5.3.0 release of Apptools. The plan is to tag the merge commit from the eventual squash-merge of this PR.

@flongford Would you have bandwidth to review?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,52 @@
 Apptools CHANGELOG
 ==================
 
+Version 5.3.0
+~~~~~~~~~~~~~
+
+Released: 2024-07-05
+
+This is a minor release, focusing primarily on bringing apptools up to date
+with respect to more recent versions of Python and 3rd party packages.
+Support for Python 3.7 has been dropped.
+
+Thanks to the following contributors:
+
+* Mark Dickinson
+* Stewart Ferguson
+* Frank Longford
+* Tony Ni
+* Sai Rahul Poruri
+
+Fixes
+-----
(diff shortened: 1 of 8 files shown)


flongford (approved):
LGTM, though with the disclaimer that I'm not very familiar with the current release process.

Nice to see how the `scriv` snippets contribute to the release notes!

mdickinson:
@flongford Thanks for review!

I've just discovered that there _are_ release notes for 5.2.1 on the 5.2.1 tag, but that they weren't ever merged back into the main branch (and they have an oddly incorrect release date). I've done my best to reconcile those with the ones introduced in this PR.

> I'm not very familiar with the current release process

There isn't really one, or at least not one that I'd care to imitate. We had a somewhat over-elaborate process for previous releases, and I'm doing something simpler this time around (in particular, since this is a minor release rather than a patch release, I'm releasing straight from the main branch). I'll add some notes on the process to the (currently empty) wiki, for possible re-use next time.

> Nice to see how the `scriv` snippets contribute to the release notes!

Yep, this was our own home-grown approximation to `scriv` (from a time that predated `scriv`s existence), introduced by @kitchoi. Now that `scriv` exists, I think we should probably switch to using that from this point onwards.

mdickinson:
> I'll add some notes on the process to the (currently empty) wiki

Started some notes here: https://github.com/enthought/apptools/wiki/Releasing

PR #362, opened 2025-05-12
Attempt to run H5 tests on macOS Silicon again

### LINKED ISSUE #344: Test workflows fail on macOS due to PyTables
The full test suite fails on macos-latest runners because PyTables does not have wheels available for Apple Silicon, and because building from an sdist fails.

We've temporarily updated the workflows in #339 to skip the HDF5-related tests on macos-latest. Those changes can be reverted when the upstream issue is addressed.

Upstream issue: https://github.com/PyTables/PyTables/issues/1165

mdickinson:
If successful, closes #344.

### DIFF
--- a/.github/workflows/test-with-pip.yml
+++ b/.github/workflows/test-with-pip.yml
@@ -20,14 +20,8 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


mdickinson:
I seem to be failing on the workflow expression syntax here - we're running the wrong branch on macOS / Python 3.9.

PR #363, opened 2025-05-13
Update changelog and version number in preparation for 5.4.1 release

mdickinson:
This PR updates the changelog and bumps the version number in preparation from the 5.4.1 release.

We're releasing from the main branch, since there are no feature-level changes since 5.4.0.


mdickinson:
Sorry: branch name is incorrect - the release should be 5.3.1, not 5.4.1.

